PR #2, opened 2018-11-23
[WIP] Carousal component

monty2610:


vinodloha:
@monty2610 any progress here?

monty2610:
Hi @vinodloha I have already updated the test cases. Please review.

lib/components/molecules/Carousal/Carousal.js
```diff
@@ -0,0 +1,103 @@
...
+
+  render() {
+    const { children, className, showDots, showArrows } = this.props;
+    const { current } = this.state;
+    const totalSlides = Children.count(children);
+    if (totalSlides === 0) {
+      return null;
+    }
+    return (
+      <section className={className}>
```
vinodloha: Don't think section would be right wrapper tag for it.

lib/components/molecules/Carousal/Carousal.js
```diff
@@ -0,0 +1,103 @@
...
+              onClick={this.moveSlideToRight}
+              disabled={current === totalSlides - 1}
+              className="arrow right"
+            >
+              {'>'}
+            </Button>
+          )}
+        </div>
+        {showDots && (
+          <Dots
```
vinodloha, line 89: How can someone configure images, instead of dots as indicators? There should be a way to do that, The dots could be default behaviour

lib/components/molecules/Carousal/Carousal.story.js
```diff
@@ -0,0 +1,100 @@
...
+    },
+    {
+      title: 'Primary Carousal Variations',
+      sections: [
+        {
+          title: 'Carousal',
+          sectionFn: () => (
+            <div style={{ width: '480px', margin: '0 auto' }}>
+              <Carousal {...defaultConfig}>
+                <Image src="https://via.placeholder.com/480.png/FF0000/fff" />
```
vinodloha, line 42: How is responsive behaviour is going to work? For example: Single slide on mobile and multiple slide to be combined of needed on large viewport.

PR #4, opened 2018-12-06
Feature/rollup based build system

vinodloha:
Now dist contains UMD, CJS and ES bundles. 

Todo: 
1. Export theme and CSS. 
2. Export other files.
3. Publish script for Nexus.


package.json
```diff
@@ -116,6 +133,20 @@
     "styled-components": "^3.4.9",
     "yup": "^0.26.6"
   },
+  "peerDependencies": {
+    "@babel/runtime": ">= 7.1.2",
+    "classnames": ">= 2.2.6",
+    "flexboxgrid": ">= 6.3.1",
+    "formik": ">= 1.3.1",
+    "next": ">= 7.0.1",
```
ankurkakroo2: What is the need for next in this setup ?

vinodloha: The Anchor Component uses Next/Link but we can take it out.

.travis.yml
```diff
@@ -0,0 +1,27 @@
...
+sudo: false
+before_install:
+- curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.9.4
+- export PATH=$HOME/.yarn/bin:$PATH
+cache:
+  yarn: true
+  directories:
+  - node_modules
+script:
+- yarn run lint
```
ankurkakroo2: why not include lint and test always in the build task

package.json
ankurkakroo2: We should take it out. React link should be fine as we are already including react as a peerDependency

PR #6, opened 2019-01-30
popover initial commit

khus29:
Initial commit. Functional test cases yet to be done.

### DIFF
--- a/lib/components/molecules/Popover/Popover.js
+++ b/lib/components/molecules/Popover/Popover.js
@@ -1,0 +1,103 @@
+// @flow
+/**
+ *props detailed explanation
+ *className - class for the popover wrapper
+ *children - child nodes of the wrapper
+ *body - The popover content
+ *preferPlace - Sets a preference of where to position the Popover.
+    valid values
+      ->above|right|below|left: prefer an explicit side
+      ->row|column: prefer an orientation
+      ->start|end: prefer an order
+      ->null - default, no preference, automatic resolution
+  *place - ????
+  *onOuterAction  - callback that executes every time user does an action
+  *refreshIntervalMs  - time bet each poll in ms. Way to position target in DOM
+  *enterExitTransitionDurationMs - time in ms that it takes to complete enter and exit animation default 500
+  *tipSize : size of tip pointer, default 7
+  *target  - React element to which popover orient itself
+  *appendTarget - Where popover will mount into. Default is 'document.body'
+*/
+import type { Node } from 'react';
+import React, { Component } from 'react';
+import styled from 'styled-components';
+import Popover from 'react-popover';
+import styles from './Popover.style';
+
+type Props = {
+  className?: string,
+  children: Node,
+  body?: Node,
+  isOpen: boolean,
+  preferPlace: string | null,
+  place: string | null,
+  onOuterAction: (SyntheticEvent<>) => void,
+  refreshIntervalMs: number,
+  enterExitTransitionDurationMs: number,
+  tipSize?: number,
+  target: Node,
+  appendTarget: Node,
+};
+type State = {
+  popoverIsOpen: boolean,
+};
+
+class PopoverWrapper extends Component<Props, State> {
+  static defaultProps = {
+    className: '',
+    body: 'test',
+    tipSize: 5,
+  };
+
+  state = {
+    popoverIsOpen: false,
+  };
+
+  props: Props;
+
+  togglePopover() {
+    this.setState(prevState => ({ popoverIsOpen: !prevState.popoverIsOpen }));
+  }
+
+  render() {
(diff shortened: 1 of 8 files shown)


lib/components/molecules/Popover/Popover.js
```diff
@@ -0,0 +1,103 @@
...
+      place,
+      onOuterAction: () => this.togglePopover(),
+      body,
+      refreshIntervalMs,
+      enterExitTransitionDurationMs,
+      tipSize,
+      target,
+      appendTarget,
+    };
+    const targetToggleProps = {
```
monty2610, line 88: Hi @khus29 this props seems to be reduntant, directly add onClick on span instead.

lib/components/molecules/Popover/Popover.js
```diff
@@ -0,0 +1,103 @@
...
+      onClick: () => this.togglePopover(),
+    };
+
+    return (
+      <Popover className={className} {...popoverProps}>
+        <span {...targetToggleProps}>{children}</span>
```
monty2610, line 94: Hi @khus29 if children can be any html, then span will be simantically wrong. Instead use block level element.

lib/components/molecules/Popover/Popover.style.js
```diff
@@ -0,0 +1,13 @@
+import { css } from 'styled-components';
+
+export default css`
+  span {
+    cursor: pointer;
+    text-decoration: none;
```
monty2610, line 6: why we need this?

package.json
```diff
@@ -128,6 +128,7 @@
     "normalize.css": "^8.0.0",
     "react": "^16.5.2",
     "react-app-polyfill": "^0.1.3",
+    "react-popover": "^0.5.10",
```
monty2610, line 131: Hi @khus29 react-popover seems to be using react portals and with react portals one issue is handling keyboard tabbing. Please check if react-popover supports tabbing.

lib/components/molecules/Popover/Popover.js
```diff
@@ -0,0 +1,103 @@
...
+  *target  - React element to which popover orient itself
+  *appendTarget - Where popover will mount into. Default is 'document.body'
+*/
+import type { Node } from 'react';
+import React, { Component } from 'react';
+import styled from 'styled-components';
+import Popover from 'react-popover';
+import styles from './Popover.style';
+
+type Props = {
```
vinodloha, line 27: Looks like you have not being following the generator for building this component, because in case of complex molecule we usually define flow types under types/index

lib/components/molecules/Popover/Popover.js
```diff
@@ -0,0 +1,103 @@
...
+    className: '',
+    body: 'test',
+    tipSize: 5,
+  };
+
+  state = {
+    popoverIsOpen: false,
+  };
+
+  props: Props;
```
vinodloha, line 56: You don't need to define this again.

lib/components/molecules/Popover/Popover.js
```diff
@@ -0,0 +1,103 @@
...
+  enterExitTransitionDurationMs: number,
+  tipSize?: number,
+  target: Node,
+  appendTarget: Node,
+};
+type State = {
+  popoverIsOpen: boolean,
+};
+
+class PopoverWrapper extends Component<Props, State> {
```
vinodloha, line 45: Should be PureComponent

lib/components/molecules/Popover/Popover.js
```diff
@@ -0,0 +1,103 @@
...
+      refreshIntervalMs,
+      enterExitTransitionDurationMs,
+      tipSize,
+      target,
+      appendTarget,
+      className,
+      children,
+    } = this.props;
+    const { popoverIsOpen } = this.state;
+    const popoverProps = {
```
vinodloha, line 76: Why define all the props again, under different name popoverProps? If you don't need to strategically place each prop, just spread it using rest operator. 
You can do: 

Const { children, ...others} = this.props; and Just use ...others instead of popoverProps.

lib/components/molecules/Popover/Popover.js
```diff
@@ -0,0 +1,103 @@
...
+
+class PopoverWrapper extends Component<Props, State> {
+  static defaultProps = {
+    className: '',
+    body: 'test',
+    tipSize: 5,
+  };
+
+  state = {
+    popoverIsOpen: false,
```
vinodloha, line 53: You are not modifying the State anywhere, so its just a prop for the react-popover. So its better if you just treat it like a Prop only and keep the component stateless.

vinodloha (changes requested):

vinodloha:
This is being done by @swapnesh now.

PR #9, opened 2019-03-15
Tooltip component created

ssi114:


vinodloha:
@ssi114 Can you check why CI build is failing and fix it?

lib/components/atoms/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,55 @@
...
+
+  showTooltip = () => {
+    this.setState({ displayTooltip: true });
+  };
+
+  render() {
+    const { message, position, className, children } = this.props;
+    const { displayTooltip } = this.state;
+    return (
+      <span className={className} onMouseLeave={this.hideTooltip}>
```
vinodloha: @ssi114 Do you think Span is the right tag for building tooltip?

lib/components/atoms/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,55 @@
...
+        {displayTooltip && (
+          <div className={classNames('tooltip-bubble', position)}>
+            <div className="tooltip-message">{message}</div>
```
vinodloha, line 60: @ssi114 i would expect tool tip to get triggered from existing components like Button

vinodloha (changes requested):
Doesn't take care of web accessibility:
https://www.w3.org/TR/wai-aria-practices-1.1/#tooltip  
https://www.sarasoueidan.com/blog/accessible-tooltips/

vinodloha (changes requested):
@ssi114 Please fix these review comments.

lib/components/atoms/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,61 @@
...
+import { Component } from 'react';
+import styled from 'styled-components';
+import classNames from 'classnames';
+
+import styles from './Tooltip.style';
+import type { Props, State } from './types';
+
+class Tooltip extends Component<Props, State> {
+  static defaultProps = {
+    message: '',
```
vinodloha, line 16: This Props will not be string.

lib/components/atoms/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,61 @@
...
+
+  hideTooltip = () => {
+    this.setState({ displayTooltip: false });
+  };
+
+  showTooltip = () => {
+    this.setState({ displayTooltip: true });
+  };
+
+  render() {
```
vinodloha, line 46: You should use react portal so that tooltip doesn't get cropped because of overflow hidden on parent.

lib/components/atoms/Tooltip/tests/Tooltip.test.js
```diff
@@ -0,0 +1,15 @@
...
+
+import Tooltip from '../index';
+
+describe('<Tooltip />', () => {
+  let TooltipComponent = '';
+  beforeEach(() => {
+    TooltipComponent = shallow(<Tooltip>Test</Tooltip>);
+  });
+
+  test('should render correctly', () => {
```
vinodloha, line 12: This test will not be enough

vinodloha (changes requested):
There are still lot of things to fix @ssi114

lib/components/atoms/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,75 @@
...
+ *
+ */
+import { Component } from 'react';
+import styled from 'styled-components';
+import classNames from 'classnames';
+
+import styles from './Tooltip.style';
+import type { Props, State } from './types';
+
+class Tooltip extends Component<Props, State> {
```
vinodloha, line 14: @ssi114 this component can be PureComponent and not a Component

lib/components/atoms/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,75 @@
...
+    className: '',
+    tabIndex: '',
+  };
+
+  state: State = {
+    displayTooltip: false,
+  };
+
+  componentDidMount() {
+    document.addEventListener('keyup', (e: KeyboardEvent) => {
```
vinodloha, line 27: keyup event should change to keydown and event type could be KeyboardEvent use SyntheticEvent instead it will cover both Mouse and Keyboard

lib/components/atoms/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,75 @@
...
+        onBlur={this.hideTooltip}
+        onFocus={this.showTooltip}
+      >
+        {displayTooltip && (
+          <div className={classNames('tooltip-bubble', position)}>
+            <div className="tooltip-message">{message}</div>
+          </div>
+        )}
+        <span className="tooltip-trigger" role={ariaRole}>
+          {children}
```
vinodloha, line 64: Children should be body of tooltip and not the trigger

sohonisaurabh:
@vinodloha I am picking up this task

PR #13, opened 2019-04-02
Feature/popover

swapnesh:
EDIT 02 -

@vinodloha Please find new changes as requested -

1. Added BEM naming convention in writing classnames.
2. Used prop based styling as suggested rather than creation of new nodes.
3.  Changed code structure to use the component body to set the Popover content and defined prop for supplying heading and trigger element/content as suggested.
4. Did bare minimum css.
5. Keeping trigger as "React$Element<*>" instead on Node as Node type is incompatible with the elements if cloned 
`const element = trigger && React.cloneElement(trigger, { onClick: this.handleClick });`
6. Removed unnecessary dummy title Helloooooo from story.js file. 

Let me know on case of any further query.

EDIT 01 -

@vinodloha Please find changes made after changing Atom to Molecule Popover. Please let me know in case of further query.

1. Changed Atom to molecule.
2. Added isVisible property to have more controlled component.
3. Added `List` component to show in the popover body component.
4. Done with bare minimum CSS part by adding border/colour etc.
5. Changed header to h3 tag than div to be more semantically correct.
6. Added close button in Popover.
7. Changed data types as requested.

New Screenshot -

Popover-molecule

---------------------------------------------------------------- 

Initial commit for Atom | Popover.

 
storybook


lib/components/atoms/Popover/Popover.js
```diff
@@ -0,0 +1,74 @@
+// @flow
```
vinodloha: @swapnesh Popover should be a molecule, Since it usually holds and displays other components. Please restructure it.

lib/components/atoms/Popover/Popover.js
```diff
@@ -0,0 +1,74 @@
...
+import type { Props, State } from './types';
+
+class Popover extends PureComponent<Props, State> {
+  static defaultProps = {
+    popOverBody: '',
+    popOverHeader: '',
+  };
+
+  state: State = {
+    showPopover: false,
```
vinodloha: Use isVisible property instead.

lib/components/atoms/Popover/Popover.mock.js
```diff
@@ -0,0 +1,7 @@
+const defaultConfig = {
```
vinodloha: Can you also make a sample where we are putting another component like LIst inside popover?

lib/components/atoms/Popover/Popover.style.js
```diff
@@ -0,0 +1,5 @@
+import { css } from 'styled-components';
+
+export default css`
+  ${props => (props.inheritedStyles ? props.inheritedStyles : '')};
```
vinodloha: Please do right bare minimum CSS like: Border Box, Small padding to all 4 side of the content.

lib/components/atoms/Popover/Popover.js
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  render() {
+    const { popOverBody, popOverHeader, children } = this.props;
+    const { showPopover } = this.state;
+    return (
+      <div role="presentation" onClick={this.handleClick} ref={this.wrapperRef}>
+        {children}
+        {showPopover && (
+          <div role="dialog">
```
vinodloha: Pop overs usually have a pointer too can we include that conditionally? Example: https://codepen.io/esaramago/pen/jbXZaz and https://getbootstrap.com/docs/4.1/components/popovers/

Also Placing a class on popOverHeader and popOverBody divs would give better styling control to people using it.

lib/components/atoms/Popover/Popover.js
```diff
@@ -0,0 +1,74 @@
...
+            <div>{popOverHeader}</div>
```
vinodloha: Semantically DIV doesn't seem right tag for it. May be a H3 will help??

lib/components/atoms/Popover/Popover.js
```diff
@@ -0,0 +1,74 @@
...
+    });
+  };
+
+  render() {
+    const { popOverBody, popOverHeader, children } = this.props;
+    const { showPopover } = this.state;
+    return (
+      <div role="presentation" onClick={this.handleClick} ref={this.wrapperRef}>
+        {children}
+        {showPopover && (
```
vinodloha: Can we also have an optional Close button with-in the popover?

lib/components/atoms/Popover/tests/Popover.test.js
```diff
@@ -0,0 +1,15 @@
+import { shallow } from 'enzyme';
+// import renderer from 'react-test-renderer';
+
+import Popover from '../index';
+
+describe('<Popover />', () => {
```
vinodloha: The test could be a bit better, at least test the code you wrote, like close and open functionality. Or any other new feature you add. Just snapshot is not enough.

lib/components/atoms/Popover/types/index.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+import type { Node } from 'react';
+
+export type Props = { popOverBody: String, popOverHeader: String, children: Node };
```
vinodloha: This is not how we define type String --> string 
I would recommend you change 
popOverHeader --> heading (It should be both string and Node type)
popOverBody --> trigger (This should be a Node or String that will trigger popover)
children --> it should be the content or body of the popover

vinodloha (changes requested):

lib/components/molecules/Popover/types/index.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+import type { Node } from 'react';
+
+export type Props = {
+  popOverBody: Node | string,
+  popOverHeader: Node | string,
+  children: React$Element<*>,
```
vinodloha: For Children correct type is Node

lib/components/molecules/Popover/Popover.story.js
```diff
@@ -0,0 +1,73 @@
...
+  noCloseBtnPopover,
+} from './Popover.mock';
+
+// Import Styled Component to showcase variations
+import Popover, { PopoverVanilla } from '.';
+
+storiesOf('Molecules', module).addWithChapters('Popover', {
+  chapters: [
+    {
+      title: `Helloooooo`,
```
vinodloha: Please Put formal Title

lib/components/molecules/Popover/Popover.story.js
```diff
@@ -0,0 +1,73 @@
...
+            allowPropTablesToggling: true,
+          },
+        },
+      ],
+    },
+    {
+      title: 'Popover Variations',
+      sections: [
+        {
+          title: '[isVisible=true] variation',
```
vinodloha: Title Should be contextual and description: Please name this as "By Default Open Variation"

lib/components/molecules/Popover/Popover.story.js
```diff
@@ -0,0 +1,73 @@
...
+          sectionFn: () => (
+            <Popover {...defaultVisiblePopover}>
+              <button>Click to Open</button>
+            </Popover>
+          ),
+        },
+        {
+          title: '[isVisible=false] variation',
```
vinodloha: By Default Close Variation

lib/components/molecules/Popover/Popover.story.js
```diff
@@ -0,0 +1,73 @@
...
+          sectionFn: () => (
+            <Popover {...hiddenPopover}>
+              <button>Click to Open</button>
+            </Popover>
+          ),
+        },
+        {
+          title: 'List(Component) variation',
```
vinodloha: List Content Variation

amrit07 (commented):
popover should be accessible via keyboard as well. Ref: https://reactjs.org/docs/accessibility.html
Also they should also breakout from a parent with overflow hidden; for e.g carousel. Ref: https://reactjs.org/docs/portals.html

lib/components/atoms/Popover/Popover.js
swapnesh: @vinodloha I made the changes as mentioned. Kindly review

lib/components/molecules/Popover/Popover.story.js
```diff
@@ -0,0 +1,73 @@
...
+            <Popover {...ListPopover}>
+              <button>Click to Open</button>
+            </Popover>
+          ),
+        },
+        {
+          title: 'Optional close button variation',
+          sectionFn: () => (
+            <Popover {...noCloseBtnPopover}>
+              <button>Click to Open</button>
```
vinodloha: See this is not really right, the content which is less is placed in the body of component (being picked as children) and the content which really is body gets set as prop. This is a wrong design pattern and would always confuse react developers. Can you please change it, i had given this comment earlier too. 

So use the component body to set the Popover content and defined prop for supplying heading and trigger element/content.

vinodloha: See this is not really right, the content which is less is placed in the body of component (being picked as children) and the content which really is body gets set as prop. This is a wrong design pattern and would always confuse react developers. Can you please change it, i had given this comment earlier too. 

So use the component body to set the Popover content and defined prop for supplying heading and trigger element/content.

vinodloha (changes requested):

lib/components/molecules/Popover/Popover.style.js
```diff
@@ -0,0 +1,18 @@
+import styled from 'styled-components';
+
+const PopoverContentWrapper = styled.div`
```
vinodloha: This is not the right way, just encapsulate all CSS in a class and apply that to whole component. Its easier that way to make variations.

lib/components/molecules/Popover/types/index.js
```diff
@@ -0,0 +1,15 @@
...
+  isVisible: boolean,
+  wrapperBackground: string,
```
vinodloha: wrapperBackground is useless, if someone wants to overwrite this component's style its not going to be background only. We should make way for enabling them to overwrite all CSS. Set class names in BEM terminology.

We are using this technique ${props => (props.inheritedStyles ? props.inheritedStyles : '')}; in all components. Please follow the same.

lib/components/molecules/Popover/tests/Popover.test.js
```diff
@@ -0,0 +1,15 @@
...
+
+import Popover from '../index';
+
+describe('<Popover />', () => {
+  let PopoverComponent = '';
+  beforeEach(() => {
+    PopoverComponent = shallow(<Popover>Test</Popover>);
+  });
+
+  test('should render correctly', () => {
```
vinodloha, line 16: Test cases needs to improve, you have a open/close functionality too. Test it here!

vinodloha (changes requested):
Please fix these review comments, there are some fundamental problems so can't accept it in current form.

lib/components/molecules/Popover/Popover.story.js
swapnesh: @vinodloha Changed the structuring as requested. Now the inner part of component represents the body of popover and trigger prop will represent the node for triggering the body content.

lib/components/molecules/Popover/types/index.js
swapnesh: Removed the unnecessary class and prop for css and aligned class naming with BEM.

lib/components/molecules/Popover/types/index.js
swapnesh: Initially I defined children as node (in new use case - its trigger prop) but getting Linting error. On github flow 
 issue thread I found that if an element is cloned then it should be mentioned as React$Element<*>

vinodloha (approved):
Looking good now! I have added few extra tests. 

But @swapnesh Thanks for this effort, must appreciated :)

PR #31, opened 2019-04-03
Initial commit for accordian

DipshiGoyal:
@vinodloha ,as disccussed I have added the functionality of accordian. Please review.
PFA
[accordian (2).zip](https://github.com/pagesource/atomic-react-components/files/3106594/accordian.2.zip)


lib/components/molecules/Accordian/Accordian.js
```diff
@@ -0,0 +1,62 @@
+// @flow
```
amrit07, line 1: Accordian generally works in group; current implementation doesn't handle this. see https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html for an example; you can also explore deatil and summary html5 elements for the same.

lib/components/molecules/Accordian/Accordian.js
```diff
@@ -0,0 +1,62 @@
...
+            Body Content
+          </AccordianBody>
+        ) : null}
+      </React.Fragment>
+    );
+  }
+}
+
+const AccordianHeader = ({ className, onClick, children }: Props): Node => (
+  <React.Fragment>
```
amrit07: Redundant usage of Fragment. below as well

lib/components/molecules/Accordian/Accordian.js
```diff
@@ -0,0 +1,62 @@
...
+ * Accordian
+ *
+ */
+import { PureComponent } from 'react';
+import styled from 'styled-components';
+import type { Node } from 'react';
+import styles from './Accordian.style';
+import type { Props, State } from './types';
+
+class Accordian extends PureComponent<Props, State> {
```
vinodloha, line 16: @DipshiGoyal i would expect you to built this component like this https://codepen.io/amitmojumder/pen/xOKjPW

vinodloha (changes requested):
Please fix review comments, let me know if you need more info

lib/components/molecules/Accordian/Accordian.js
DipshiGoyal: The suggestion has been incorporated.

lib/components/molecules/Accordian/tests/Accordian.test.js
```diff
@@ -0,0 +1,15 @@
+import { shallow } from 'enzyme';
+// import renderer from 'react-test-renderer';
+
+import Accordian from '../index';
+
+describe('<Accordian />', () => {
```
vinodloha, line 4: The test needs to be much better, you need to test the expend collapse functionality

lib/components/molecules/Accordian/Accordian.style.js
```diff
@@ -0,0 +1,46 @@
+import styled from 'styled-components';
+
+const AccordianWrapper = styled.div`
+  display: block;
+  width: 50%;
+  margin: 10px 0;
+`;
+
+const AccordianHeader = styled.div`
```
vinodloha: This is not correct style of applying CSS, you can follow same BEM style in styled components too.

lib/components/molecules/Accordian/Accordian.js
```diff
@@ -0,0 +1,46 @@
...
+    const doesShow = this.state.isOpen;
+    this.setState({ isOpen: !doesShow });
+  };
+
+  render() {
+    const { accordianHeader, accordianBody, className } = this.props;
+
+    return (
+      <AccordianWrapper>
+        <AccordianHeader
```
vinodloha: AccordianHeader will not be DIV but a summary tag, you should built it as separate component.  Same applies to Detail too. You can export them same as accordion and people can loop over them and build accordion groups.

vinodloha:
@DipshiGoyal lets connect today, i feel there is some gap in our understanding for this component.

PR #54, opened 2019-09-28
404 Links inside index.md file

singhtal:
Fixed all Github links inside index.md file going 404 due to double occurrence of "readme" in their url's.


vinodloha (approved):
Thanks for fixing this @tal1992

PR #57, opened 2019-10-04
tooltip review points implemented

monikajoon:
Tooltip component


lib/components/molecules/ToolTip/Tooltip.js
```diff
@@ -0,0 +1,36 @@
...
+
+const Tooltip = ({ tootTipContent, domElement, showImage, ...otherProps }: Props): Div => {
+  return (
+    <div
+      className={otherProps.className}
+      showImage={showImage}
+      {...otherProps}
+      data-tooltip={tootTipContent}
+    >
+      {domElement}
```
vinodloha: You should be using children Pattern here.

lib/components/molecules/ToolTip/Tooltip.js
```diff
@@ -0,0 +1,36 @@
...
+ */
+import React from 'react';
+import styled from 'styled-components';
+
+import styles from './Tooltip.style';
+import type { Props, Div } from './types';
+
+const Tooltip = ({ tootTipContent, domElement, showImage, ...otherProps }: Props): Div => {
+  return (
+    <div
```
vinodloha: Where is the trigger element for this tooltip? and also the close button? Sorry but looks like an incomplete implementation

vinodloha (changes requested):
Looks like an incomplete implementation.

lib/components/molecules/ToolTip/Tooltip.js
monikajoon: Have implemented hover and it is in style file.

lib/components/molecules/ToolTip/Tooltip.js
```diff
@@ -0,0 +1,81 @@
+import React, { PureComponent } from 'react';
+
+import type { Props, State } from './types';
+import Image from '../../atoms/Image';
+import TooltipContainer from './TooltipContainer';
+
+class Tooltip extends PureComponent<Props, State> {
+  state = {
+    isOpen: false,
+    style: {},
```
vinodloha: Why you need style in state?

lib/components/molecules/ToolTip/Tooltip.js
```diff
@@ -0,0 +1,81 @@
...
+    if (isOpen && element && !element.contains(event.target)) {
+      this.setState({ isOpen: false });
+    }
+  };
+
+  render() {
+    const { isOpen, style } = this.state;
+    const { title, domElement, image, tootTipContent } = this.props;
+    return (
+      <div ref={this.toolTipElement}>
```
vinodloha: Can you also take care of basic web accessibility for tooltips. https://pressbooks.library.ryerson.ca/wafd/chapter/tooltips/

Also implement this via portals. By default it will not have a wrapper div.

monikajoon (commented):
@vinodloha , Please review. Implemented review points.

monikajoon:
@vinodloha , Please review this. Implemented Review points.

PR #73, opened 2019-10-14
issue-72 | HOC plop generator file updated

bhardwajamit0890:


generators/component/index.js
```diff
@@ -16,7 +16,7 @@ module.exports = {
       name: 'type',
       message: 'Select the type of component',
       default: 'Stateless Function',
-      choices: () => ['Stateless Function', 'React.Component'],
+      choices: () => ['Stateless Function', 'React.Component', 'hoc'],
```
vinodloha: Let's keep the option as HOC

vinodloha (changes requested):
Do this small change please

generators/component/hoc.js.hbs
```diff
@@ -0,0 +1,12 @@
+// @flow
+/**
+ *
+ * {{ properCase name }}
+ *
+ */
+
+const {{ properCase name}} = Component => {
```
vinodloha: This template is not enough. Have minimum imports. I.e React, types definitions etc. 
Rename Component as WrappedComponent.

Template should looks like below: 
```
import React from 'react';

import getDisplayName from '../../core/utils/getDisplayName';

const {{ properCase name}} = WrappedComponent => {
  // ...and returns another component...
  class extends React.Component {
    constructor(props) {
      super(props);
      this.state = {};
    }

    render() {
    
      return <WrappedComponent {...this.props} />;
    }
  };
  {{ properCase name}}.displayName = `{{ properCase name}}(${getDisplayName(WrappedComponent)})`;
  return {{ properCase name}};
}
``` 

Also create a file at core/utils with following content. 
function getDisplayName(WrappedComponent) {
  return WrappedComponent.displayName || WrappedComponent.name || 'Component';
}

generators/utils/componentExists.js
```diff
@@ -11,9 +11,11 @@ const atomComponents = fs.readdirSync(path.join(__dirname, '../../lib/components
 const moleculeComponents = fs.readdirSync(path.join(__dirname, '../../lib/components/molecules'));
 const organismComponents = fs.readdirSync(path.join(__dirname, '../../lib/components/organisms'));
 const templateComponents = fs.readdirSync(path.join(__dirname, '../../lib/components/templates'));
+const hoc = fs.readdirSync(path.join(__dirname, '../../lib/components/hoc'));
```
vinodloha, line 14: I believe we are also creating test/* dir and files too.

vinodloha (changes requested):

bhardwajamit0890:
Done

PR #116, opened 2019-11-08
issue-72 | updated class name for hoc

bhardwajamit0890:


vinodloha:
Can you give a bit of details about this Bug? Looks like it was not letting you define the variable again, is that true?

vinodloha (approved):

PR #118, opened 2019-11-12
issue-72 | using hooks for HOC class component

bhardwajamit0890:


generators/component/hoc.js.hbs
```diff
@@ -5,23 +5,24 @@
...
 const {{ camelCase name}} = WrappedComponent => {
   // ...and returns another component...
-  class {{ properCase name}} extends React.Component {
-    constructor(props) {
-      super(props);
-      this.state = {};
-    }
 
-    render() {
+  function {{ properCase name}}(props) {
```
bhardwajamit0890: Need to add props and node return type.

vinodloha (approved):

PR #134, opened 2019-11-28
File-Upload: Molecule Component

salilkapoor:


lib/components/molecules/Form/FieldFileInput/FieldFileInput.js
```diff
@@ -0,0 +1,84 @@
...
+  name,
+  label,
+  onChange,
+  onBlur,
+  labelProps,
+  fieldProps,
+  ...others
+}: Props): Node => (
+  // TODO: Add support for label position
+  <div className={className}>
```
vinodloha: Does it support drag and drop based upload in a container? That was one of the req too

lib/components/molecules/Form/FieldFileInput/tests/FieldFileInput.test.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { shallow, render } from 'enzyme';
+import { FieldFileInput } from '../FieldFileInput';
+
+describe('FieldInput Component', () => {
```
vinodloha, line 6: These tests are good start but not enough, ask your self does it cover every line of code you have written in component?

lib/components/molecules/Form/Form.js
```diff
@@ -10,6 +10,7 @@ import FieldSelect from './FieldSelect';
 import FieldTextarea from './FieldTextarea';
 import FieldRadioGroup from './FieldRadioGroup';
 import FieldCheckbox from './FieldCheckbox';
+import FieldFileInput from './FieldFileInput';
```
vinodloha, line 13: Nice

lib/components/molecules/Form/FieldFileInput/FieldFileInput.js
```diff
@@ -0,0 +1,84 @@
...
+    <Field
+      id={id}
+      name={name}
+      type={type}
+      render={({ field, form }) => {
+        const { handleChange, handleBlur } = form;
+        return (
+          <Input
```
ankurkakroo2: Think of it this way.
Where is the CSS for input really residing ? 
FieldInput is just a wrapper and should have CSS only for label  and outer styling.

You have 2 options:
Write CSS in Input.style.js for [type=file]

or better: create a new component named FileInput, Use <Input type="file"> inside it and write css in FileInput.style.js

And then wrap everything in this component

Reach out to me if you need further clarification.

lib/components/molecules/Form/FieldFileInput/FieldFileInput.js
```diff
@@ -0,0 +1,84 @@
+// @flow
```
ankurkakroo2, line 1: Can you attach screenshots in your PR ?

lib/components/molecules/Form/FieldFileInput/FieldFileInput.style.js
```diff
@@ -0,0 +1,18 @@
+import { css } from 'styled-components';
+
+// TODO: Fix the duplicate styling in input.style.js
+const styles = css`
+  margin-bottom: 0.75rem;
```
ankurkakroo2: We should be using margins and paddings from variables

lib/components/molecules/Form/FieldFileInput/FieldFileInput.js
salilkapoor: image

lib/components/molecules/Form/FieldFileInput/FieldFileInput.style.js
salilkapoor: I changed to variabled for colors, spacing wherever applicable.

lib/components/molecules/Form/FieldFileInput/FieldFileInput.js
salilkapoor: I followed second approach as suggested. Also, it was giving me problem in one more case i.e. value field as well, when i tried uploaded file and it was not able to set explicit value prop.

lib/components/molecules/Form/FieldFileInput/FieldFileInput.js
salilkapoor: Implemented.

PR #137, opened 2019-11-28
Carousel Feature

jainaman23:


lib/components/atoms/ListItem/ListItem.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ */
+import React from 'react';
+import styled from 'styled-components';
+
+import styles from './ListItem.style';
+import type { Props } from './types';
+
+const ListItem = ({ className, children }: Props): Node => {
+  const classes = ['item', className].join(' ');
```
vinodloha: You don't need to do this we already have a module classnames that does this class joining well

lib/components/atoms/ListItem/ListItem.js
```diff
@@ -0,0 +1,27 @@
...
+ * ListItem
+ *
+ */
+import React from 'react';
+import styled from 'styled-components';
+
+import styles from './ListItem.style';
+import type { Props } from './types';
+
+const ListItem = ({ className, children }: Props): Node => {
```
vinodloha, line 14: Why is this component not a child of list or carousel. We can't use it without List

lib/components/atoms/ListItem/ListItem.story.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+// import { action } from '@storybook/addon-actions';
+import defaultConfig from './ListItem.mock';
+
+// Import Styled Component to showcase variations
+import ListItem, { ListItemVanilla } from '.';
+
+// TODO: indentation is not working fine in editor. Need to check for .hbs files
+// $FlowFixMe eslint-disable-next-line extra-rules/no-commented-out-code
```
vinodloha: Remove this line

lib/components/atoms/ListItem/ListItem.story.js
```diff
@@ -0,0 +1,19 @@
...
+storiesOf('Atoms', module)
+  .addParameters({ jest: ['ListItem', 'ListItemVanilla'] })
+  .add('CTA Knobs', () => (
```
vinodloha: This the component title

lib/components/atoms/ListItem/tests/ListItem.test.js
```diff
@@ -0,0 +1,14 @@
+import { shallow } from 'enzyme';
+
+import ListItem from '../index';
+
+describe('<ListItem />', () => {
```
vinodloha, line 5: Test is not enough

lib/components/molecules/AtomicCarousel/AtomicCarousel.js
```diff
@@ -0,0 +1,129 @@
...
+import { ButtonVanilla } from '../../atoms/Button';
+
+// Rendering all dots as button
+const getSliderDots = (slidesCount, currentItem, setItem) => {
+  const specificNode = useCallback(index => setItem(index));
+  const countElements = [];
+
+  for (let i = 1; i <= slidesCount; i += 1) {
+    countElements.push(
+      <ListItem key={i} className="atomic-item">
```
vinodloha: Using I for key is wrong

lib/components/molecules/AtomicCarousel/AtomicCarousel.js
```diff
@@ -0,0 +1,129 @@
...
+        </ButtonVanilla>
+      </ListItem>
+    );
+  }
+
+  return countElements;
+};
+
+// Calculating required number of dots
+const getList = (items, setItem, currentItem, perSlideFrame) => {
```
vinodloha, line 36: Where does slide count come from? Why can't you find this by children.length

lib/components/molecules/AtomicCarousel/AtomicCarousel.js
```diff
@@ -0,0 +1,129 @@
...
+    } else {
+      itemsArray.push(content[currentScreen - 1]);
+    }
+
+    return itemsArray;
+  };
+
+  return (
+    <>
+      <article className={carouselClasses}>
```
vinodloha: Do you really think article is the right tag for carousel?

lib/components/molecules/AtomicCarousel/AtomicCarousel.js
```diff
@@ -0,0 +1,129 @@
...
+        {navigation && (
+          <ButtonVanilla className="previous" onClick={handlePrev} disabled={currentSlide === 1}>
+            {prevLabel}
+          </ButtonVanilla>
+        )}
+        {data.length !== 0 && getPerFrameItems(data, perSlideFrame, currentSlide)}
+        {navigation && (
```
vinodloha, line 116: What if i want to position the both buttons in same place?

lib/components/molecules/AtomicCarousel/AtomicCarousel.js
```diff
@@ -0,0 +1,129 @@
...
+          <ButtonVanilla
+            className="next"
+            onClick={handleNext}
+            disabled={Math.ceil(data.length / perSlideFrame) === currentSlide}
+          >
+            {nextLabel}
+          </ButtonVanilla>
+        )}
+      </article>
+      {dots && getList(data, setCurrentSlide, currentSlide, perSlideFrame)}
```
vinodloha, line 126: I believe these Dots too are part of Carousel component, why are they being placed inside?

lib/components/molecules/AtomicCarousel/AtomicCarousel.js
```diff
@@ -0,0 +1,129 @@
...
+// Rendering all dots as button
+const getSliderDots = (slidesCount, currentItem, setItem) => {
+  const specificNode = useCallback(index => setItem(index));
+  const countElements = [];
+
+  for (let i = 1; i <= slidesCount; i += 1) {
+    countElements.push(
+      <ListItem key={i} className="atomic-item">
+        <ButtonVanilla disabled={currentItem === i} onClick={() => specificNode(i)}>
+          {i}
```
vinodloha: What if i want to place an Image thumb instead of an Image?

lib/components/molecules/AtomicCarousel/AtomicCarousel.js
```diff
@@ -0,0 +1,129 @@
...
+  navigation,
+  children,
+  className,
+  perSlideFrame,
+  prevLabel,
+  nextLabel,
+}: Props): Node => {
+  const carouselClasses = [...className.split(','), 'atomic-carousel'].join(' ');
+  const [currentSlide, setCurrentSlide] = useState(1);
+  const [data, setData] = useState([]);
```
vinodloha, line 57: What does data and setData mean? What does it represent

lib/components/molecules/AtomicCarousel/AtomicCarousel.js
```diff
@@ -0,0 +1,129 @@
...
+
+  return (
+    <>
+      <article className={carouselClasses}>
+        {navigation && (
+          <ButtonVanilla className="previous" onClick={handlePrev} disabled={currentSlide === 1}>
+            {prevLabel}
+          </ButtonVanilla>
+        )}
+        {data.length !== 0 && getPerFrameItems(data, perSlideFrame, currentSlide)}
```
vinodloha, line 115: This is not a declarative style of working

lib/components/molecules/AtomicCarousel/AtomicCarousel.story.js
```diff
@@ -0,0 +1,35 @@
...
+import defaultConfig from './AtomicCarousel.mock';
+
+// Import Styled Component to showcase variations
+import AtomicCarousel, { AtomicCarouselVanilla } from '.';
+
+// TODO: indentation is not working fine in editor. Need to check for .hbs files
+// $FlowFixMe eslint-disable-next-line extra-rules/no-commented-out-code
+storiesOf('Molecules', module)
+  .addParameters({ jest: ['AtomicCarousel', 'AtomicCarouselVanilla'] })
+  .add('CTA Knobs', () => (
```
vinodloha: Rename the story

lib/components/molecules/AtomicCarousel/AtomicCarousel.style.js
```diff
@@ -0,0 +1,21 @@
+import { css } from 'styled-components';
+
+const styles = css`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+
+  & > * + * {
```
vinodloha: seriously??

lib/components/molecules/AtomicCarousel/tests/AtomicCarousel.test.js
```diff
@@ -0,0 +1,43 @@
...
+  });
+
+  it('render without dots', () => {
+    const wrapper = shallow(
+      <AtomicCarousel dots={false}>
+        <h1>Test</h1>
+      </AtomicCarousel>
+    );
+    expect(wrapper.props().dots).toBe(false);
+    expect(wrapper.find('.atomic-dots').length).toBe(0);
```
vinodloha, line 26: To query this you need render and not shallow are you sure this is working?

lib/components/molecules/AtomicCarousel/tests/AtomicCarousel.test.js
```diff
@@ -0,0 +1,43 @@
+import { shallow, mount } from 'enzyme';
+
+import AtomicCarousel from '../index';
+
+describe('<AtomicCarousel />', () => {
```
vinodloha, line 5: These tests are not enough

lib/components/molecules/AtomicCarousel/AtomicCarousel.js
```diff
@@ -0,0 +1,129 @@
...
+
+  useEffect(() => {
+    if (children !== undefined && !Array.isArray(children)) {
+      setData([children]);
+    } else {
+      setData(children);
+    }
+  }, []);
+
+  const getPerFrameItems = (content, perFrame, currentScreen) => {
```
vinodloha, line 92: Its impossible to understand the propose of these methods

lib/components/molecules/AtomicCarousel/types/index.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+
+export type Props = {
+  dots: Boolean,
```
vinodloha, line 4: Why is everything a mandatory prop

lib/components/molecules/AtomicCarousel/AtomicCarousel.js
jainaman23: It is being calculated by using perSlideFrame and children's length.

lib/components/molecules/AtomicCarousel/AtomicCarousel.js
jainaman23: I can also use section, was confused with section and article as both are the same in some respect so I decided to go with the article as they can be used as an independent identity.

lib/components/molecules/AtomicCarousel/AtomicCarousel.js
jainaman23: It's the default/vanilla/traditional design. So if you need to change the order we can do with order property as parent is flex.

lib/components/molecules/AtomicCarousel/AtomicCarousel.js
jainaman23: Sometimes we need dots to be at the custom position. It would be easy to make dots position custom with CSS. and It is the part of the carousel but not of the slider section.

lib/components/molecules/AtomicCarousel/AtomicCarousel.js
jainaman23: data is the actual slider content as this component is using children prop so it could be a single element or multiple elements and when there is a single element child return object and when there are multiple it returns the array,

So to make the consistency I have used data state and convert children into this array.

lib/components/molecules/AtomicCarousel/types/index.js
jainaman23: Didn't get this. Doesn't it need to be isRequired to make it mandatory?

PR #146, opened 2019-11-29
Add files via upload

maliksagar85:


App.css
```diff
@@ -0,0 +1,46 @@
...
+  margin-top: 5px;
+}
+
+.footer-heading{
+  font-weight: bold;
+  padding-bottom: 5px;
+}
+
+.footer-content{
+  padding-top: 15px;
```
vinodloha, line 28: Why no shorthand properties

App.css
```diff
@@ -0,0 +1,46 @@
+footer {
```
vinodloha, line 1: You had to style the components based on the theme we had setup as per design system

Footer.jsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+import _ from 'lodash';
+import $ from 'jquery';
```
vinodloha, line 3: Jquery in React application

vinodloha (changes requested):
This is one of the most poorly done exercise

PR #150, opened 2019-12-01
Accessibility fixed for input

bhardwajamit0890:


config/jest.setup.js
```diff
@@ -1,4 +1,9 @@
 const { configure } = require('enzyme');
 const Adapter = require('enzyme-adapter-react-16');
 
+const lodash = require.requireActual('lodash');
+
 configure({ adapter: new Adapter() });
+lodash.uniqueId = () => {
+  return 1;
```
vinodloha: This is not good enough mock.

lib/components/molecules/Form/FieldInput/FieldInput.js
```diff
@@ -60,13 +66,21 @@ const FieldInput = ({
           );
         }}
       </Field>
+      {validationInstructions ? (
+        <Para id={`${inputID}-instruction`} className="hide-default-sample">
```
vinodloha: className="hide-default-sample" class is just for storybook. Let this text be visible

package.json
```diff
@@ -133,6 +133,7 @@
     "core-js": "^2.6.5",
     "flexboxgrid": "^6.3.1",
     "formik": "^2.0.6",
+    "lodash": "^4.17.15",
```
manishekhawat, line 136: if we are introducing lodash only for this utility of uniqueId across the application, then lets write our own logic to generate uniqueId or just copy that particular utility method. @vinodloha your thoughts?

ankurkakroo2: @bhardwajamit0890  Lodash has modularised their code a lot. 
Just use: https://www.npmjs.com/package/lodash.uniqueid
I know what you have done will also result in a tree-shaken code, but still ;)

vinodloha (approved):

PR #152, opened 2019-12-03
Feature/react table

ankush29:
Create a column atom just for testing table children. please ignore it.
pending: sorting and have to test for every atom component would be rendered in it as child


lib/components/atoms/Column/Column.js
```diff
@@ -0,0 +1,22 @@
...
+ * Column
+ *
+ */
+import React from 'react';
+import styled from 'styled-components';
+
+import styles from './Column.style';
+import type { Props } from './types';
+
+const Column = ({ intent, label, value }) =>
```
vinodloha: From a Column are you referring to a table cell? 

Also the way you are using it you will end up putting a span in each cell. Which is a big overhead of nodes in table context. Will it not be nice if this component knows when to return <TH> or When just <TD> 

We should better name it a Cell because thats what it is. When you export it as static prop, you can use Column.

lib/components/atoms/Column/Column.story.js
```diff
@@ -0,0 +1,19 @@
...
+import defaultConfig from './Column.mock';
+
+// Import Styled Component to showcase variations
+import Column, { ColumnVanilla } from '.';
+
+// TODO: indentation is not working fine in editor. Need to check for .hbs files
+// $FlowFixMe eslint-disable-next-line extra-rules/no-commented-out-code
+storiesOf('Atoms', module)
+  .addParameters({ jest: ['Column', 'ColumnVanilla'] })
+  .add('CTA Knobs', () => (
```
vinodloha: This is not a CTA Knobs story, please change accordingly.

lib/components/atoms/Column/tests/Column.test.js
```diff
@@ -0,0 +1,14 @@
+import { shallow } from 'enzyme';
+
+import Column from '../index';
+
+describe('<Column />', () => {
```
vinodloha: Again!!! You are still not developing components as TDD

lib/components/molecules/RenderTable/RenderTable.js
```diff
@@ -0,0 +1,51 @@
...
+import styles from './RenderTable.style';
+import type { renderTableProps } from './types';
+
+const RenderTable = ({ children, data, className, ...props }: renderTableProps) => {
+  return (
+    <table className={className} {...props}>
+      <thead>
+        <tr>
+          {Children.map(children, column => (
+            <th key={column.props.name}>{cloneElement(column, { intent: 'header' })}</th>
```
vinodloha: First of all, the React.cloneElement example only works if your child is a single React element. 

I don't see a use-case of cloneElement, you should be using Children here

lib/components/molecules/RenderTable/RenderTable.js
```diff
@@ -0,0 +1,51 @@
+// @flow
+/**
+ *
+ * RenderTable
```
vinodloha: Can we not just name it Table?

lib/components/molecules/RenderTable/RenderTable.js
```diff
@@ -0,0 +1,51 @@
...
+        </tr>
+      </thead>
+      <tbody>
+        {data.map(item => (
+          <tr key={item.id}>
+            {Children.map(children, column => (
+              <td>
+                {cloneElement(column, {
+                  intent: 'cell',
+                  value: item[column.props.name],
```
vinodloha: Can this item be a component?

lib/components/molecules/RenderTable/RenderTable.js
```diff
@@ -0,0 +1,51 @@
...
+          {Children.map(children, column => (
+            <th key={column.props.name}>{cloneElement(column, { intent: 'header' })}</th>
+          ))}
+        </tr>
+      </thead>
+      <tbody>
+        {data.map(item => (
+          <tr key={item.id}>
+            {Children.map(children, column => (
+              <td>
```
vinodloha: Break it down to a smaller component.

lib/components/molecules/RenderTable/RenderTable.js
```diff
@@ -0,0 +1,51 @@
...
+import styled from 'styled-components';
+
+import styles from './RenderTable.style';
+import type { renderTableProps } from './types';
+
+const RenderTable = ({ children, data, className, ...props }: renderTableProps) => {
+  return (
+    <table className={className} {...props}>
+      <thead>
+        <tr>
```
vinodloha: Going by this design the table will always be Horizontal one where T-Head is at the top; What if we need T-Head as first cell of each row?

lib/components/molecules/RenderTable/RenderTable.story.js
```diff
@@ -0,0 +1,27 @@
...
+      <Column label="Dessert (100g serving)" name="name" />
+      <Column label="Calories" name="calories" />
+      <Column label="Fat (g)" name="fat" />
+      <Column label="Carbs (g)" name="carbs" />
+      <Column label="Protein (g)" name="protein" />
+    </RenderTableVanilla>
+  ))
+  .add('Table', () => (
+    <RenderTable {...defaultConfig}>
+      <Column label="Dessert (100g serving)" name="name" />
```
vinodloha: Good work, I like it. One improvement. Column should be a static props of Table so that. 

<Table.Column name="name">Dessert (100g serving)</Table.Column>

lib/components/molecules/RenderTable/tests/RenderTable.test.js
```diff
@@ -0,0 +1,81 @@
...
+    wrapper = shallow(
+      <RenderTableVanilla data={data}>
+        <Column label="Dessert (100g serving)" name="name" />
+        <Column label="Calories" name="calories" />
+        <Column label="Fat (g)" name="fat" />
+        <Column label="Carbs (g)" name="carbs" />
+        <Column label="Protein (g)" name="protein" />
+      </RenderTableVanilla>
+    );
+    table = wrapper.find('table');
```
vinodloha: Define variable were you use them so that they are local to the test and safe from tempering

package.json
```diff
@@ -114,6 +114,7 @@
     "mini-css-extract-plugin": "^0.8.0",
     "plop": "^2.4.0",
     "prettier": "1.19.1",
+    "react-test-renderer": "^16.12.0",
```
vinodloha: There was't really need for it. Its is part of enzyme-adapter-react-16

lib/components/molecules/RenderTable/RenderTable.js
```diff
@@ -0,0 +1,51 @@
...
+ * RenderTable
+ *
+ */
+import React, { cloneElement, Children } from 'react';
+import styled from 'styled-components';
+
+import styles from './RenderTable.style';
+import type { renderTableProps } from './types';
+
+const RenderTable = ({ children, data, className, ...props }: renderTableProps) => {
```
vinodloha: How will you support cellpadding, cellspacing and other such table props ?

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,59 @@
...
+import styles from './Table.style';
+// eslint-disable-next-line flowtype/no-types-missing-file-annotation
+import type { Props, ColumnProps } from './types';
+
+// eslint-disable-next-line flowtype/no-types-missing-file-annotation
+const Table = ({ children, data, className, ...props }: Props) => {
+  return (
+    <table className={className} {...props} role="table">
+      <thead>
+        <tr role="row">
```
vinodloha, line 13: Are you assuming all tables to be Horizontal with header at the top? I had asked this question on earlier PR too.

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,59 @@
...
+import styled from 'styled-components';
+
+import styles from './Table.style';
+// eslint-disable-next-line flowtype/no-types-missing-file-annotation
+import type { Props, ColumnProps } from './types';
+
+// eslint-disable-next-line flowtype/no-types-missing-file-annotation
+const Table = ({ children, data, className, ...props }: Props) => {
+  return (
+    <table className={className} {...props} role="table">
```
vinodloha: What the point of role="table" on a table element?

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,59 @@
...
+      <thead>
+        <tr role="row">
+          {Children.map(children, column => (
+            <th key={column.props.name} role="columnheader">
```
vinodloha, line 15: key={column.props.name} are you imagining every cell to have a name prop?

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,59 @@
...
+        </tr>
+      </thead>
+      <tbody>
+        {data.map(item => (
+          <tr key={item.id} role="row">
+            {Children.map(children, column => (
+              <td role="cell">
+                {cloneElement(column, {
+                  intent: 'cell',
+                  value: item[column.props.name],
```
vinodloha: How is this cell different from Header Cell, both can render plain text or components

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,59 @@
...
+            ))}
+          </tr>
+        ))}
+      </tbody>
+    </table>
+  );
+};
+
+Table.defaultProps = {
+  className: '',
```
vinodloha: There is no point of empty values in default props

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,59 @@
...
+  intent: '',
+};
+
+// eslint-disable-next-line flowtype/no-types-missing-file-annotation
+Table.Column = ({ intent, label, value }: ColumnProps) => {
+  return intent === 'header' ? <em>{label}</em> : <em>{value}</em>;
```
vinodloha: Seriously?? Every table cell will wrap its content in <em>? For header its still imaginable but for TD doesn't make sense

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,59 @@
...
+};
+
+Table.Column.defaultProps = {
+  intent: '',
```
vinodloha: Same why empty Default props

lib/components/molecules/Table/Table.story.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import defaultConfig from './Table.mock';
+import Table, { TableVanilla } from '.';
+
+storiesOf('Molecules', module)
+  .addParameters({ jest: ['Table', 'TableVanilla'] })
+
+  .add('Table Knobs', () => (
```
vinodloha, line 12: Where is a test case and sample for table being able to hold existing elements in cell?

lib/components/molecules/Table/Table.style.js
```diff
@@ -0,0 +1,13 @@
+import { css } from 'styled-components';
+
+export default css`
+  border-spacing: ${props => props.theme.spacing.unit0};
```
vinodloha: The variable used is wrong, UNIT_1 is the connect format and starts from 1. Please check spacing.js under theme

lib/components/molecules/Table/Table.style.js
```diff
@@ -0,0 +1,13 @@
...
+  ${props => props.theme.border.FIELD_SECONDARY};
+  th,
+  td {
+    margin: 0;
+    padding: 0.5rem;
```
vinodloha, line 9: Why use hard coded values?

lib/components/molecules/Table/tests/Table.test.js
```diff
@@ -0,0 +1,122 @@
...
+
+  it('should have th is equal to number of columns', () => {
+    expect(headers.length).toEqual(5);
+  });
+
+  it('should have tr is equal to number of data rows', () => {
+    expect(rows).toHaveLength(data.length);
+  });
+
+  it('should loop through each row and check the content', () => {
```
vinodloha, line 63: How do you know if your cell are containing right data as per your table headings

lib/components/molecules/Table/Table.story.js
```diff
@@ -0,0 +1,26 @@
...
+      <TableVanilla.Column label="Dessert (100g serving)" name="name" />
+      <TableVanilla.Column label="Calories" name="calories" />
+      <TableVanilla.Column label="Fat (g)" name="fat" />
+      <TableVanilla.Column label="Carbs (g)" name="carbs" />
+      <TableVanilla.Column label="Protein (g)" name="protein" />
+    </TableVanilla>
+  ))
+  .add('Table', () => (
+    <Table {...defaultConfig}>
+      <Table.Column label="Dessert (100g serving)" name="name" />
```
vinodloha: I love this, but still...

what would be better if you don't take the value from property and use the body to allow the heading to be defined from outside. 

The prop name should be named more clearly so we know it maps the cell to its data.

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,59 @@
...
+            <th key={column.props.name} role="columnheader">
+              {cloneElement(column, { intent: 'header' })}
+            </th>
+          ))}
+        </tr>
+      </thead>
+      <tbody>
+        {data.map(item => (
+          <tr key={item.id} role="row">
+            {Children.map(children, column => (
```
vinodloha, line 29: Don't like this nested loop, we should have broken this complicity by creating a smaller component.

lib/components/molecules/Table/types/index.js
```diff
@@ -0,0 +1,22 @@
+// @flow
+import type { Node } from 'react';
+
+type Content =
```
vinodloha: Will your table always accept these props only? Its content you can't define that with component since use-cases can be limitless

lib/components/molecules/Table/types/index.js
```diff
@@ -0,0 +1,22 @@
...
+export type Props = {
+  className?: string,
+  data: Array<Content>,
+  children: Node,
+};
+
+export type ColumnProps = {
+  intent?: string,
+  label?: string,
+  value?: string | number,
```
vinodloha, line 12: It could be a Node too

vinodloha (dismissed):
Still lot of improvements required and far away from being complete

PR #153, opened 2019-12-04
Feature/design system basics

vinodloha:
- Setup Design System Tokens 
- Applied Design Tokens in Styles for all Components 
- Fixed the breaking test cases with Enzyme mocks


.lintstagedrc
```diff
@@ -3,9 +3,14 @@
...
+    "yarn run lint:css:fix",
+  ],
   "*.js": [
     "yarn run lint:css:fix",
-    "yarn run lint:js:fix",
-    "yarn run test -- --findRelatedTests",
-  ]
+    "yarn run lint:js:fix"
+  ],
+  "*.test.js": [
```
ankurkakroo2: If we are doing this, then the test cases should run when JS files are changed as well

lib/styles/themes/base/def/index.js
```diff
@@ -0,0 +1,24 @@
...
+import behaviour from './behaviour';
+import border from './border';
+import colors from './colors';
+import dimensions from './dimensions';
+import typography from './typography';
+import zIndex from './zIndex';
+import shadow from './shadow';
+import spacing from './spacing';
+
+export default {
```
ankurkakroo2, line 14: Why not expose them separately. 

The way you will be eventually consuming this will be 

theme.spacing.something else which may not be ideal. 
These are JS imports, if you don't want animations and behaviour, you anyways get them as a packaged deal.


We can talk about it

lib/styles/themes/base/def/spacing.js
```diff
@@ -0,0 +1,17 @@
...
+const MAX_SPACING_UNITS = 20;
+const spacingGenerator = () => {
+  const spacing = {};
+  let remLastUnit = 0;
+  const remIncrementUnit = 0.5; // Equal to 8 pixel
+  for (let i = 1; i <= MAX_SPACING_UNITS; i += 1) {
+    remLastUnit += remIncrementUnit;
+    spacing[`unit${i}`] = `${remLastUnit}rem`;
+  }
+  spacing.quadUnit = '0.125rem';
```
ankurkakroo2: change it to caps

lib/components/atoms/Anchor/Anchor.style.js
```diff
@@ -2,22 +2,14 @@ import { css } from 'styled-components';
 
 export default css`
   cursor: pointer;
-  text-decoration: none;
-  color: ${props => props.theme.linkColorPrimary};
+  ${props => props.theme.behaviour.LINK_DECORATION};
```
ankurkakroo2, line 5: Drilling down the props may result in a cumbersome experience. 
2 ways to minimise: either send the reqd values separately or club multiple props in one arrow function

.lintstagedrc
vinodloha: Ah, you are right.

lib/components/atoms/Anchor/Anchor.style.js
vinodloha: I know it looks repetitive but this is just going to produce CSS. I find this style much leaner.

ankurkakroo2:
@vinodloha  Build failing

lib/styles/themes/base/def/index.js
vinodloha: Not much to gain, all of this get consumed in CSS. Also ThemeProvider accepts a theme object, we will have to combine it there and maintain the configuration of passing whole theme entities in rest of the uses.

lib/styles/themes/base/def/spacing.js
vinodloha: Done

PR #155, opened 2019-12-04
feature/video

ritugupta64:


.vscode/launch.json
```diff
@@ -0,0 +1,21 @@
+{
```
vijendramisra, line 1: Please remove this file from commit and add this in ignore list.

lib/components/atoms/Video/Video.js
```diff
@@ -0,0 +1,60 @@
...
+      height={height}
+      controls={controls}
+      autoPlay={autoPlay}
+      loop={loop}
+      muted={muted}
+      poster={poster}
+      className={className}
+      controlsList={controlsList}
+    >
+      <source src={src} type={`video/${src.substring(src.lastIndexOf('.') + 1)}`} />
```
vijendramisra, line 40: Can't we take video type also as a prop instead of calculating this from video URL?

lib/components/atoms/Video/Video.story.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+// import { action } from '@storybook/addon-actions';
```
vijendramisra: Remove commented code

lib/components/atoms/Video/Video.story.js
```diff
@@ -0,0 +1,39 @@
...
+          kind="subtitles"
+          srcLang={caption.srcLang}
+          label={caption.label}
+        />
+      ))}
+    </VideoVanilla>
+  ))
+  .add('Video', () => (
+    <Video {...defaultConfig}>
+      {' '}
```
vijendramisra: remove this extra space.

lib/components/atoms/Video/Video.js
ritugupta64: Yes, we can do but don't want to do any redundant. So to give the more flexibility we can get TYPE from URL.

.vscode/launch.json
ritugupta64: added

ritugupta64: added in ignore list

lib/components/atoms/Video/Video.story.js
ritugupta64: Thanks, removed

lib/components/atoms/Video/Video.story.js
ritugupta64: Thanks, removed

.eslintrc
```diff
@@ -41,7 +41,12 @@
     ],
     "prettier/prettier": "error",
     "compat/compat": "error",
-    "linebreak-style": "off"
+    "linebreak-style": "off",
+    "jsx-a11y/media-has-caption": [ 1, {
```
vinodloha, line 45: I thought we managed to fix this by putting a default Track

lib/components/atoms/Track/Track.js
```diff
@@ -0,0 +1,23 @@
...
+ *
+ */
+import React from 'react';
+import styled from 'styled-components';
+
+import styles from './Track.style';
+import type { Props } from './types';
+
+const Track = ({ src, srcLang, id, label, isDefault }: Props): Node => (
+  <track src={src} default={isDefault} kind="subtitles" srcLang={srcLang} label={label} key={id} />
```
vinodloha, line 14: Why kind attribute is hard coded? Make use of default props

lib/components/atoms/Track/Track.js
```diff
@@ -0,0 +1,23 @@
...
+);
+
+Track.defaultProps = {};
+
+export default styled(Track)`
```
vinodloha, line 19: Does Track need CSS styles?

lib/components/atoms/Track/tests/Track.test.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import { TrackVanilla } from '../index';
+
+describe('<Track />', () => {
```
vinodloha, line 6: These tests are not enough

lib/components/atoms/Track/types/index.js
```diff
@@ -0,0 +1,6 @@
+// @flow
+import type { Node } from 'react';
+
+export type Props = {
```
vinodloha, line 4: Why only this Prop? Please define your props here that components receives

lib/components/atoms/Video/Video.js
```diff
@@ -0,0 +1,63 @@
...
+      {children.length > 0 ? children : <Track {...defaultCaption} />}
+    </video>
+  );
+};
+
+Video.defaultProps = {
+  width: '',
```
vinodloha, line 47: Use a default value, most likely all your videos be 100% to their container. May be its better to preserve a aspect ratio through CSS.

lib/components/atoms/Video/Video.js
```diff
@@ -0,0 +1,63 @@
...
+  height: '',
+  autoPlay: false,
+  controls: true,
+  loop: false,
+  muted: false,
+  poster: '',
+  src: 'http://techslides.com/demos/sample-videos/small.mp4',
+  controlsList: '',
```
vinodloha, line 55: There is no point in setting '' value as default. Make it optional if you have to

lib/components/atoms/Video/Video.mock.js
```diff
@@ -0,0 +1,19 @@
+const defaultConfig = {
+  width: '100%',
```
vinodloha: Please mark optional and mandatory props clearly.

lib/components/atoms/Video/tests/Video.test.js
```diff
@@ -0,0 +1,34 @@
...
+  it('should have source tag', () => {
+    expect(VideoComponent.find('video').find('source')).toBeTruthy();
+  });
+
+  it('should have src attribute in source tag', () => {
+    const VideoComponentRender = render(<VideoVanilla {...defaultConfig} />);
+    expect(VideoComponentRender.find('source').attr('src')).toEqual(defaultConfig.src);
+  });
+
+  it('should be autoplay but need to be mute', () => {
```
vinodloha, line 24: There could be a situation where the config passed had { autoplay: false }, do test that behaviour too

We must test all use cases our components can render?

lib/components/atoms/Video/types/index.js
```diff
@@ -0,0 +1,18 @@
...
+  controls: Boolean,
+  loop?: Boolean,
+  muted?: Boolean,
+  poster?: String,
+  src: String,
+  type: String,
+  children?: Node,
+  className?: String,
+  controlsList?: String,
+  defaultCaption: Object,
```
vinodloha, line 17: shouldn't defaultCaption be also optional?

vinodloha (changes requested):
Still away from being merge worthy

.eslintrc
ritugupta64: yes, removed and this fixed has been done in other branch 162

lib/components/atoms/Track/Track.js
ritugupta64: Thanks, done in separate branch #162

lib/components/atoms/Track/Track.js
ritugupta64: removed..no need

lib/components/atoms/Track/types/index.js
ritugupta64: done in separate

lib/components/atoms/Video/types/index.js
ritugupta64: yes, done

lib/components/atoms/Video/Video.js
ritugupta64: correct, i HAVE REMOVED THE default props for src but keeping here muted as false so that anyone have understanding to update it

lib/components/atoms/Video/Video.js
ritugupta64: as we are not passing any value for width So, width will be automatically expandable with their ancestor So no need to write any css for this particular container. 
If there will any need to add the css then we have class prop and then we can decide the values there.

lib/components/atoms/Track/tests/Track.test.js
ritugupta64: yes working,

PR #156, opened 2019-12-06
Implemented carousel with test cases.

jainaman23:


jainaman23:
Not valid branch name

PR #159, opened 2019-12-06
Feature/accordion

aditimalhotra19:
Working on TDD


lib/components/molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,41 @@
...
+	accordionContent? : [],
+	isOpen? : boolean,
+};
+
+const Accordion = ({ className, accordionContent, isOpen }: Props): Node => 
+	<div className={className}>
+		{accordionContent && accordionContent.map((item, index) => (
+			<details className="accordion-item" key={`accordion-${index.toString()}`} open={item.isOpen}>
+				<summary className="accordion-header">
+					{item.accordionHeader}
```
vinodloha: This is still not good, the components in react should make use of declarative way of definition. 

To do this Panel and Header should be independent components

lib/components/molecules/Accordion/Accordion.story.js
```diff
@@ -0,0 +1,43 @@
...
+
+// Import Styled Component to showcase variations
+import Accordion, {
+    AccordionVanilla
+} from '.';
+
+storiesOf('Molecules', module).addParameters({
+    jest: ['Accordion', 'AccordionVanilla']
+}).addWithChapters('Accordion', {
+    chapters: [{
```
vinodloha: This is an old API we don't use anymore, please take latest from Develop

PR #161, opened 2019-12-09
'DateSelector'

kirtisharma1:


lib/components/atoms/Select/Select.js
```diff
@@ -7,7 +7,7 @@ import styles from './Select.style';
 type Props = {
   id: string,
   className?: string,
-  name: string,
+  name?: string,
```
vinodloha: Name should be mandatory

lib/components/atoms/Select/Select.js
```diff
@@ -37,7 +37,7 @@ const Select = ({
   >
     {placeholder && <option value="">{placeholder}</option>}
     {options.map(opt => (
-      <option key={opt} value={opt}>
+      <option key={`${name}_${opt}`} value={opt}>
```
vinodloha: Nice, instead of ${opt} use index that will avoid white space

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,203 @@
...
+import React, { useEffect, useState } from 'react';
+import styled from 'styled-components';
+
+import classNames from 'classnames';
+import styles from './DateSelector.style';
+import type { Props } from './types';
+import Select from '../../atoms/Select';
+import Label from '../../atoms/Label';
+
+const monthOptions = [
```
vinodloha: How will these change if locale changes to lets say French or Spanish

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,203 @@
...
+  'July',
+  'August',
+  'September',
+  'October',
+  'November',
+  'December',
+];
+
+const DateSelector = (props: Props): Node => {
+  const { className, format, startDate, endDate } = props;
```
vinodloha: Move this line at like 31

(props: Props) --> { className, format, startDate, endDate } will achieve same thing

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,203 @@
...
+  const [selectedDate, setSelectedDate] = useState(new Date());
+  const [newStartDate, setNewStartDate] = useState(
```
vinodloha: How is startDate and newStartDate diff?

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,203 @@
...
+    new Date(new Date().setFullYear(new Date().getFullYear() - 100))
+  );
+  const [newEndDate, setNewEndDate] = useState(
+    new Date(new Date().setFullYear(new Date().getFullYear() + 50))
```
vinodloha: Where did you find this 50 or 100 number?

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,203 @@
...
+  );
+
+  const setDate = dateType => {
+    if (dateType === 'start') {
+      if (startDate) {
+        setNewStartDate(new Date(startDate));
+      } else {
+        const d = new Date();
+        setNewStartDate(new Date(d.setFullYear(d.getFullYear() - 100)));
```
vinodloha: new Date(d.setFullYear(d.getFullYear() - 100) get repeated too often, create a method for it

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,203 @@
...
+  const [selectedDate, setSelectedDate] = useState(new Date());
+  const [newStartDate, setNewStartDate] = useState(
+    new Date(new Date().setFullYear(new Date().getFullYear() - 100))
+  );
+  const [newEndDate, setNewEndDate] = useState(
+    new Date(new Date().setFullYear(new Date().getFullYear() + 50))
+  );
+
+  const setDate = dateType => {
+    if (dateType === 'start') {
```
vinodloha: Such else and If cases produce duplicate code, the same function should be able to handle or set start or end date 

the old diff is + and - that can be managed with a small ? : ;

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,203 @@
...
+  useEffect(() => {
+    setSelectedDate(newStartDate);
+  }, [newStartDate, newEndDate]);
+
+  const getDaysInMonth = () => {
+    const noOfDays = new Date(selectedDate.getFullYear(), selectedDate.getMonth(), 0).getDate();
+    const days = [];
+    let startDay = 1;
+    let endDay = noOfDays;
+    if (
```
vinodloha: I don't understand what this logic below is doing, first line itself has given you all the values. 

You must break this complexity

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,203 @@
...
+    ) {
+      startDay = newStartDate.getDate();
+    }
+    if (
+      selectedDate.getMonth() === newEndDate.getMonth() &&
+      selectedDate.getFullYear() === newEndDate.getFullYear()
+    ) {
+      endDay = newEndDate.getDate();
+    }
+    for (let i = startDay; i <= endDay; i += 1) {
```
vinodloha: Hope you know end date and start date are optional features so that might be required or might not be in some use-cases

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,203 @@
...
+      days.push(i);
+    }
+    return days;
+  };
+
+  const getMonthOptions = () => {
+    let months = [...monthOptions];
+    if (selectedDate.getFullYear() === newStartDate.getFullYear()) {
```
vinodloha: So much of duplication

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,203 @@
...
+  const getYearOptions = () => {
+    const years = [];
+    for (let i = newStartDate.getFullYear(); i <= newEndDate.getFullYear(); i += 1) {
+      years.push(i);
+    }
+    return years;
+  };
+
+  const DaySelector = (
+    <Label htmlFor="day">
```
vinodloha: Where is text for label? don't use the label to wrap the field set relation ship with for htmlFor="{`${id}_day`}"

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,203 @@
...
+      <Select
+        onChange={e => {
+          return setSelectedDate(
+            new Date(selectedDate.getFullYear(), selectedDate.getMonth(), e.target.value)
+          );
+        }}
+        id="day"
```
vinodloha: Don't hard code id, create one from the one passed for whole component

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,203 @@
...
+        name="day"
+        className={classNames('date', className)}
+        disabled={false}
```
vinodloha: disabled={false} ??? hard coding

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,203 @@
...
+        elementLocator="date-picker-year"
+        selectedOption={selectedDate.getFullYear()}
+        options={getYearOptions()}
+      />
+    </Label>
+  );
+
+  return (
+    <div className={className}>
+      {(format.toLowerCase() === 'ddmmyy' || format.toLowerCase() === 'ddmmyyyy') && [
```
vinodloha: Prefer switch cases

lib/components/molecules/DateSelector/types/index.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+import type { Node } from 'react';
+
+type Format = 'ddmmyy' | 'ddmmyyyy' | 'mmddyy' | 'mmddyyyy' | 'mmyy' | 'mmyyyy';
+
+export type Props = {
+  children: Node,
+  className?: string,
+  format?: Format,
+  startDate: number,
```
vinodloha: startDate and endDate are also optional

vinodloha (changes requested):
Looks like a good progress but still far way from being complete

lib/components/atoms/Select/Select.js
```diff
@@ -50,6 +50,7 @@ Select.defaultProps = {
   placeholder: '',
   className: '',
   elementLocator: '',
+  name: 'optionName',
```
vinodloha: Name should not be part of default props also don't like how we are setting empty default values, please fix it too

lib/components/molecules/DateSelector/DateSelector.js
vinodloha: I think we can get this dynamically, from browser itself and then we will no longer need to take care of locales

vinodloha (changes requested):
Please work on review comments

lib/components/atoms/Select/Select.js
kirtisharma1: Updated

lib/components/atoms/Select/Select.js
kirtisharma1: updated

lib/components/molecules/DateSelector/DateSelector.js
kirtisharma1: implemented

lib/components/molecules/DateSelector/DateSelector.js
kirtisharma1: updated

lib/components/molecules/DateSelector/DateSelector.js
kirtisharma1: removed the startDate from state

lib/components/molecules/DateSelector/DateSelector.js
kirtisharma1: Those were for the default start and end date

lib/components/atoms/Select/Select.js
kirtisharma1: Updated.

lib/components/molecules/DateSelector/types/index.js
kirtisharma1: Updated

lib/components/molecules/DateSelector/DateSelector.js
kirtisharma1: Implemented

lib/components/molecules/DateSelector/DateSelector.js
kirtisharma1: Updated

lib/components/molecules/DateSelector/DateSelector.js
kirtisharma1: Updated

lib/components/molecules/DateSelector/DateSelector.js
kirtisharma1: these are optional features but to show default start and end dates I'm using these

lib/components/molecules/DateSelector/DateSelector.js
kirtisharma1: That was a mistake. Made changes

lib/components/molecules/DateSelector/DateSelector.js
kirtisharma1: Changed the logic

lib/components/atoms/Select/Select.js
```diff
@@ -46,10 +50,10 @@ const Select = ({
 
 Select.defaultProps = {
   disabled: false,
-  selectedOption: '',
-  placeholder: '',
-  className: '',
-  elementLocator: '',
+  selectedOption: 'select',
```
vinodloha: By default selected option should be null

lib/components/atoms/Select/Select.js
```diff
@@ -46,10 +50,10 @@ const Select = ({
...
+  placeholder: 'Select',
+  className: 'select',
+  elementLocator: 'select',
```
vinodloha: Set elementLocator default value as null

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,184 @@
...
+    const daysOptions = [];
+    let startDay = 1;
+    let endDay = new Date(selectedDate.getFullYear(), selectedDate.getMonth() + 1, 0).getDate();
+    if (
+      selectedDate.getMonth() === startDate.getMonth() &&
+      selectedDate.getFullYear() === startDate.getFullYear()
+    ) {
+      startDay = startDate.getDate();
+    }
+    if (
```
vinodloha: Still there is redundant code we can reduce, for example:

1.  These 2 If conditions were we are startDay and endDay could be returned by a function.

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,184 @@
...
+    }
+    for (let i = startDay; i <= endDay; i += 1) {
+      daysOptions.push(i);
+    }
+    return daysOptions;
+  };
+
+  const getMonthOptions = () => {
+    let monthOptions = [...months];
+    if (selectedDate.getFullYear() === startDate.getFullYear()) {
```
vinodloha: Same here,  this could be returned by a function and also clear case of If or else.

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,184 @@
...
+    if (selectedDate.getFullYear() === endDate.getFullYear()) {
+      const endMonth = endDate.getMonth();
+      monthOptions = monthOptions.slice(0, endMonth + 1);
+    }
+    return monthOptions;
+  };
+
+  const getYearOptions = () => {
+    const years = [];
+    for (let i = startDate.getFullYear(); i <= endDate.getFullYear(); i += 1) {
```
vinodloha: Can't we generate this ES6 way?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from

https://stackoverflow.com/questions/39924644/es6-generate-an-array-of-numbers

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,184 @@
...
+        name="year"
+        className={classNames('year', className)}
+        selectedOption={selectedDate.getFullYear()}
+        options={getYearOptions()}
+      />
+    </>
+  );
+
+  const renderSelector = value => {
+    switch (value.toLowerCase()) {
```
vinodloha: Nice

lib/components/molecules/DateSelector/types/index.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+type Format = 'ddmmyy' | 'ddmmyyyy' | 'mmddyy' | 'mmddyyyy' | 'mmyy' | 'mmyyyy';
```
vinodloha: nice

vinodloha:
@kirtisharma1 please resolve Conflict. 

Also Every select box should have a label. Lets place that at top of the select box.

lib/components/molecules/DateSelector/DateSelector.js
```diff
@@ -0,0 +1,184 @@
...
+};
+
+DateSelector.defaultProps = {
+  id: 'date-selector',
+  className: 'date-selector',
+  format: 'ddmmyy',
+  startDate: new Date(),
+  endDate: new Date(new Date().setFullYear(new Date().getFullYear() + 50)),
+  locale: 'default',
+  dateLabel: 'Date',
```
vinodloha: Don't set below as default value, Make them mandatory, Remove the showLabel: false. For accessibility its important that every field has a label.  

1. dateLabel: 'Date',
2. monthLabel: 'Month',
3. yearLabel: 'Year',

PR #168, opened 2019-12-10
Feature/accordion

aditimalhotra19:
Accordion changes


lib/components/molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,51 @@
...
+import React from 'react';
+import type { Node } from 'react';
+import styled from 'styled-components';
+import styles from './Accordion.style';
+import AccordionHeader from './AccordionHeader';
+import AccordionPanel from './AccordionPanel';
+
+type Props = {
+  className?: string,
+  accordionContent?: [],
```
vinodloha: Why accordionContent has to be an array? You could achieve this by {children}

Adopt the declarative style that JSX provides

lib/components/molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,51 @@
...
+const Accordion = ({ className, accordionContent, isOpen }: Props): Node => (
+  <div className={className}>
+    {accordionContent &&
+      accordionContent.map((item, index) => (
+        <details
+          className="accordion-item"
+          key={`accordion-${index.toString()}`}
+          open={item.isOpen}
+        >
+          <AccordionHeader className={item.headerClassName || ''}>
```
vinodloha: AccordionHeader and AccordionPanel should be exposed outside as compound components so that users of accordion can define their bodies in decorative manner.

lib/components/molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,51 @@
...
+            {item.accordionHeader}
+          </AccordionHeader>
+          <AccordionPanel className={item.panelClassName || ''}>
+            {item.accordionBody}
```
vinodloha: {item.accordionBody} should have been handled with children. 

Same for {item.accordionHeader}

lib/components/molecules/Accordion/Accordion.mock.js
```diff
@@ -0,0 +1,47 @@
+const accordionContent = [
```
vinodloha: This is very bad way of using a component in React, should always be avoided.

lib/components/molecules/Accordion/AccordionHeader/tests/AccordionHeader.test.js
```diff
@@ -0,0 +1,19 @@
...
+
+import { AccordionHeaderVanilla } from '../index';
+
+describe('<AccordionHeader />', () => {
+  let AccordionHeaderComponent = '';
+  beforeEach(() => {
+    AccordionHeaderComponent = shallow(<AccordionHeaderVanilla>Test</AccordionHeaderVanilla>);
+  });
+
+  test('should render', () => {
```
vinodloha: This is not enough, where do you test the if your content is being placed properly

lib/components/molecules/Accordion/AccordionHeader/AccordionHeader.js
```diff
@@ -0,0 +1,30 @@
...
+
+import styles from './AccordionHeader.style';
+
+type Props = {
+  className?: string,
+  children: Node,
+};
+
+const AccordionHeader = ({ children, className }: Props): Node => (
+  <summary className={className}>{children}</summary>
```
vinodloha: How will you make it work in IE 11

lib/components/molecules/Accordion/AccordionPanel/tests/AccordionPanel.test.js
```diff
@@ -0,0 +1,15 @@
...
+
+import { AccordionPanelVanilla } from '../index';
+
+describe('<AccordionPanel />', () => {
+  let AccordionPanelComponent = '';
+  beforeEach(() => {
+    AccordionPanelComponent = shallow(<AccordionPanelVanilla>Test</AccordionPanelVanilla>);
+  });
+
+  test('should render correctly', () => {
```
vinodloha: Tests are not enough.

lib/components/molecules/Accordion/AccordionPanel/AccordionPanel.story.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+// import { action } from '@storybook/addon-actions';
+import defaultConfig from './AccordionPanel.mock';
+
+// Import Styled Component to showcase variations
+import AccordionPanel, { AccordionPanelVanilla } from '.';
+
+storiesOf('Molecules/AccordionPanel', module)
```
vinodloha: There is no need to have storybook config for just header and panel, we can't test these independently.

lib/components/molecules/Accordion/AccordionPanel/AccordionPanel.js
```diff
@@ -0,0 +1,30 @@
...
+import styled from 'styled-components';
+
+import styles from './AccordionPanel.style';
+
+type Props = {
+  className?: string,
+  children: Node,
+};
+
+const AccordionPanel = ({ children, className }: Props): Node => (
```
vinodloha: This could have been just a local component to Accordion component itself instead of making it a full fledge component since it can't be used independently

lib/components/molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,51 @@
...
+import AccordionPanel from './AccordionPanel';
+
+type Props = {
+  className?: string,
+  accordionContent?: [],
+  isOpen?: boolean,
+};
+
+// eslint-disable-next-line no-unused-vars
+const Accordion = ({ className, accordionContent, isOpen }: Props): Node => (
```
vinodloha: What have you done from accessibility stand point?

vinodloha (changes requested):
Still this is far from being perfect but you have come a long way @aditimalhotra19

lib/components/molecules/Accordion/AccordionPanel/tests/AccordionPanel.test.js
aditimalhotra19: Done

lib/components/molecules/Accordion/AccordionHeader/tests/AccordionHeader.test.js
aditimalhotra19: Added

lib/components/molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,49 @@
...
+  </details>
+);
+
+AccordionPanel.defaultProps = {
+  className: 'accordion-panel',
+};
+
+AccordionPanel.header = AccordionHeader;
+
+const Accordion = ({ className, children }: AccordionProps): Node => (
```
vinodloha, line 88: What have you done from Accessibility perspective?

http://accessibleculture.org/articles/2012/03/screen-readers-and-details-summary/

There should be fallback for unsupported browsers. Should work with dynamic import

lib/components/molecules/Accordion/tests/Accordion.test.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { mount } from 'enzyme';
+// eslint-disable-next-line import/named
+import { defaultAccordion } from '../Accordion.mock';
+import Accordion from '../index';
+
+describe('<Accordion />', () => {
```
vinodloha, line 5: These tests are just not enough, They are not even testing for small components that accordion uses.

lib/components/molecules/Accordion/tests/Accordion.test.js
```diff
@@ -0,0 +1,25 @@
...
+  let AccordionComponent = '';
+
+  beforeEach(() => {
+    AccordionComponent = mount(<Accordion {...defaultAccordion} />);
```
vinodloha: Is this how your storybook is using Accordion component??

vinodloha (changes requested):
Test are in poor condition

lib/components/molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,63 @@
...
+};
+
+const AccordionHeader = ({ children, className, ariaLabelAttr }: AccordionHeaderProps): Node => {
+  const accordionContext = useContext(AccordionContext);
+
+  return (
+    <summary
+      className={className}
+      role="button"
+      aria-expanded="false"
```
vinodloha: aria-expanded="false" should change to aria-expanded="true" when its in open state

lib/components/molecules/Accordion/tests/Accordion.test.js
```diff
@@ -0,0 +1,39 @@
...
+import { mount } from 'enzyme';
+// eslint-disable-next-line import/named
+import { defaultAccordion } from '../Accordion.mock';
+import Accordion from '../index';
+
+describe('<Accordion />', () => {
+  let AccordionComponent = '';
+
+  beforeEach(() => {
+    AccordionComponent = mount(
```
vinodloha, line 9: Put the content as well and test for its rendering in right places, trigger the expend collapse

lib/components/molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,69 @@
...
+ */
+import React, { useContext, useState } from 'react';
+import type { Node } from 'react';
+import styled from 'styled-components';
+import styles from './Accordion.style';
+import type { AccordionHeaderProps, AccordionPanelProps, AccordionProps } from './types';
+
+const AccordionContext = React.createContext('');
+const AccordionPanel = ({ children, className, id }: AccordionPanelProps): Node => (
+  <AccordionContext.Provider value={id}>
```
vinodloha: What? Why does AccordionContext.Provider is needed at all? Who is teaching you this rocket science?

lib/components/molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,85 @@
...
+import styles from './Accordion.style';
+import type { AccordionHeaderProps, AccordionPanelProps, AccordionProps } from './types';
+
+const AccordionHeader = ({
+  children,
+  className,
+  ariaLabelAttr,
+  accordionPanelId,
+}: AccordionHeaderProps): Node => {
+  const [ariaExpandedVal, toggleAriaExpandedOnClick] = useState(false);
```
vinodloha: ariaExpandedVal --> isExpended

This state should be maintained in the Panel Component, because that component is getting expended and collapsed. You can pass the isExpended and toggleExpended handler down to this component and it will get called on click.

lib/components/molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,85 @@
...
+const AccordionPanel = ({ children, className, id }: AccordionPanelProps): Node => {
+  const childrenWithProps = React.Children.map(children, accordionheader => {
+    if (accordionheader.type.displayName === AccordionHeader.displayName)
+      return React.cloneElement(accordionheader, { accordionPanelId: id });
+
+    return React.cloneElement(accordionheader);
+  });
+  return (
+    <details className={className} id={id}>
+      {childrenWithProps}
```
vinodloha: Create a wrapper and place the AccordionHeader and User defined Children of Panel separately in a wrapper. That wrapper can be controlled with isExpended state that you will maintain here. You can use it to hide and show the content inside the Detail's body. 

Then you will not need any ploy-fill in unsupported browsers. .

lib/components/molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,107 @@
...
+  const [isExpanded, toggleExpanded] = useState(false);
+
+  const handleClick = () => toggleExpanded(!isExpanded);
+
+  return (
+    <summary
+      className={className}
+      role="button"
+      aria-expanded={isExpanded}
+      aria-label={ariaLabelAttr}
```
vinodloha: ariaLabelAttr --> ariaLabel

lib/components/molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,107 @@
...
+      aria-labelledby={accordionPanelId}
+      onClick={handleClick}
+      onKeyPress={handleClick}
```
vinodloha: onKeyPress={handleClick} is wrong. You must handle which keys it will respond to

lib/components/molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,109 @@
...
+  return (
+    <summary
+      className={className}
+      role="button"
+      aria-expanded={isExpanded}
+      aria-label={ariaLabel}
+      aria-labelledby={accordionPanelId}
+      onClick={handleClick}
+      onKeyPress={handleOnKeyPress}
+      tabIndex="0"
```
vinodloha, line 34: You can't add tabIndex="0" by default, it will bring default focus from the page to this element. If we need to make it focusable we should conditionally set and remove it tabIndex="-1"

lib/components/molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,109 @@
...
+import styles from './Accordion.style';
+import type { AccordionHeaderProps, AccordionPanelProps, AccordionProps } from './types';
+
+const AccordionHeader = ({
+  children,
+  className,
+  ariaLabel,
+  accordionPanelId,
+}: AccordionHeaderProps): Node => {
+  const [isExpanded, toggleExpanded] = useState(false);
```
vinodloha: What you did here is functionally right but this should have happened on the AccordionPanel Component and from here you will only fire the clickHandler passed from AccordionPanel Component. The state will be passed down from AccordionPanel and this component when clicked just toggle it and send into the action to be updated in the parent component. 

Please understand the Detail is collapsing not the summary. So state should be maintained by AccordionPanel

vinodloha (changes requested):
Still some logic needs to be moved around.

lib/components/molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * AccordionPanel
+ *
+ */
+const AccordionPanel = ({ className, children, id }: AccordionPanelProps): Node => {
+  const [isHidden, toggleHidden] = useState(true);
+  const isIE = false || !!document.documentMode;
+  const [setActive, setActiveState] = useState('');
+  const toggleAccordion = () => {
+    if (isIE) setActiveState(setActive === '' ? 'active' : '');
```
vinodloha: There is no need to check the IE. You can just always toggle the attributes and the collapse/expend will be controlled with CSS

lib/components/molecules/Accordion/tests/Accordion.test.js
```diff
@@ -0,0 +1,53 @@
...
+    expect(AccordionComponent.find('summary').text()).toBe(defaultAccordion.accordionHeader);
+  });
+
+  test('check the panel content', () => {
+    expect(AccordionComponent.find('p').text()).toBe(defaultAccordion.accordionPanelContent);
+  });
+
+  test('panel opens on click of header', () => {
+    AccordionComponent.find('summary').simulate('click');
+    expect(AccordionComponent.find('summary').props()).toHaveProperty('aria-expanded', true);
```
vinodloha, line 36: Nice

lib/components/molecules/Accordion/Accordion.js
aditimalhotra19: As discussed, I validated tab indexing on accordion after adding button before it. It works fine and the tab index comes on button before accordion. Hence, keeping the change as it is.

lib/components/molecules/Accordion/Accordion.js
```diff
@@ -0,0 +1,109 @@
...
+    toggleHidden(!isHidden);
+  };
+  const childrenWithHeaderProps = React.Children.map(children, (accordionheader, i) => {
+    if (accordionheader.type.displayName === AccordionHeader.displayName)
+      return React.cloneElement(accordionheader, { accordionPanelId: id });
+
+    if (i === 1)
```
vinodloha: I think this condition should not be here since it limits you to only pick the second child. Please achieve this with renderProp

vinodloha (approved):

PR #173, opened 2019-12-11
Feature/react table

ankush29:


lib/components/molecules/InputChoice/InputChoice.mock.js
```diff
@@ -33,6 +33,14 @@ const checkboxWithLabelBeforeInput = {
   showLabelBeforeInput: true,
 };
 
+const checkboxHeader = {
```
vinodloha, line 36: Nice

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,94 @@
+/* eslint-disable flowtype/no-types-missing-file-annotation */
+import React, { cloneElement, Children } from 'react';
+import styled from 'styled-components';
+
+import styles from './Table.style';
+// eslint-disable-next-line flowtype/no-types-missing-file-annotation
+import type { Props, ColumnProps } from './types';
+
+// eslint-disable-next-line flowtype/no-types-missing-file-annotation
```
vinodloha: Put a flow declaration at the top and remove all these ignore statements

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,94 @@
...
+    </table>
+  );
+};
+
+Table.defaultProps = {
+  className: 'className',
+};
+
+// eslint-disable-next-line flowtype/no-types-missing-file-annotation
+Table.Column = ({ intent, value, render }: ColumnProps) => {
```
vinodloha: Since your table can be vertical or horizontal calling this Table.Column is not really right. Find a better name

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,94 @@
...
+          </tr>
+        ))}
+      </tbody>
+    );
+  };
+
+  const VerticalHeader = () => {
+    return Children.map(children, column => (
+      <tr key={column.props.name}>
+        <th role="columnheader">{cloneElement(column, { intent: 'header' })}</th>
```
vinodloha: Why are you using cloneElement? the same can be done with Children

Here is more explanation 
https://stackoverflow.com/questions/37521798/when-should-i-be-using-react-cloneelement-vs-this-props-children/50441271#50441271

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,94 @@
...
+          <Head />
+          <Tbody />
+        </>
+      )}
+    </table>
+  );
+};
+
+Table.defaultProps = {
+  className: 'className',
```
vinodloha: Do you think this **className** will have a **className** as class name? Common!!!

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,94 @@
...
+};
+
+// eslint-disable-next-line flowtype/no-types-missing-file-annotation
+Table.Column = ({ intent, value, render }: ColumnProps) => {
+  const { method, showInCell, showInHeader, content } = render;
+  if (showInHeader && showInCell) return method();
+  if (showInCell) return intent === 'header' ? <em>{content}</em> : method();
+  return intent === 'header' ? <em>{content}</em> : <em>{value}</em>;
```
vinodloha: I am writing this 3rd time, why you want to make use of <Em> tags each each cell. Doesn't make sense to me at all. Why not just use TH instead

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,94 @@
...
+};
+
+Table.defaultProps = {
+  className: 'className',
+};
+
+// eslint-disable-next-line flowtype/no-types-missing-file-annotation
+Table.Column = ({ intent, value, render }: ColumnProps) => {
+  const { method, showInCell, showInHeader, content } = render;
+  if (showInHeader && showInCell) return method();
```
vinodloha: You ate still not following a declarative approach. Read this below blog it will only take 5 mins

https://codeburst.io/declarative-vs-imperative-programming-a8a7c93d9ad2

lib/components/molecules/Table/Table.story.js
```diff
@@ -0,0 +1,62 @@
...
+import { checkboxHeader } from '../InputChoice/InputChoice.mock';
+import InputChoice from '../InputChoice';
+import Button from '../../atoms/Button';
+import { primaryButton } from '../../atoms/Button/Button.mock';
+
+storiesOf('Molecules', module)
+  .addParameters({ jest: ['Table', 'TableVanilla'] })
+  .add('Table Knobs', () => (
+    <TableVanilla {...defaultConfig}>
+      <Table.Column dataMappingKey="name" render={{ content: 'Dessert (100g serving)' }} />
```
vinodloha: Why are you not using the body of this component?

lib/components/molecules/Table/Table.story.js
```diff
@@ -0,0 +1,62 @@
...
+      <Table.Column dataMappingKey="fat" render={{ content: 'Fat (g)' }} />
+      <Table.Column dataMappingKey="carbs" render={{ content: 'Carbs (g)' }} />
+      <Table.Column dataMappingKey="protein" render={{ content: 'Protein (g)' }} />
+    </Table>
+  ))
+  .add('Table with input box', () => (
+    <Table {...defaultConfig}>
+      <Table.Column
+        render={{
+          showInHeader: true,
```
vinodloha: I don't see following defined as default values anywhere

showInHeader: false,
          showInCell: true,

lib/components/molecules/Table/Table.js
ankush29: done

lib/components/molecules/Table/tests/Table.test.js
```diff
@@ -0,0 +1,111 @@
...
+  it('should have tbody component', () => {
+    expect(tbody.length).toEqual(1);
+  });
+  it('should have th is equal to number of columns', () => {
+    expect(headers.length).toEqual(5);
+  });
+  it('should have tr is equal to number of data rows', () => {
+    expect(rows).toHaveLength(data.length);
+  });
+  it('should loop through each row and check the content', () => {
```
vinodloha, line 73: Nice, but no need to loop, as long as you know one row is rendering properly rest should be fine too

lib/components/molecules/Table/tests/Table.test.js
```diff
@@ -0,0 +1,111 @@
+import { shallow, mount } from 'enzyme';
+
+import { TableVanilla } from '../index';
+
+describe('Table', () => {
```
vinodloha, line 5: Tests are much better now

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,73 @@
...
+import React, { Children } from 'react';
+import styled from 'styled-components';
+
+import styles from './Table.style';
+import type { Props } from './types';
+
+const Table = ({ children, data, className, vertical, ...props }: Props) => {
+  const Head = child => (
+    <th key={child.props.name} role="columnheader">
+      {child.props.children}
```
vinodloha: There is no real need to keep the Head under Table. It can just be a small functional component in same file but outside and then you will not need to do this {child.props.children} instead of {children}

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,73 @@
...
+    </th>
+  );
+
+  const Body = (child, item) => {
```
vinodloha: Move it out

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,73 @@
...
+    const { dataMappingKey } = child.props;
+    return (
+      <td role="cell">
+        {typeof dataMappingKey === 'function' ? dataMappingKey() : item[dataMappingKey]}
+      </td>
+    );
+  };
+
+  const VerticalHeader = child => {
```
vinodloha: Move it out as independent component, also this should be a component like Cell, why does it matter if it renders a cell for vertical or horizontal

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,73 @@
...
+    return data.map(item => (
+      <td role="cell" key={item.id}>
+        {item[child.props.dataMappingKey]}
+      </td>
+    ));
+  };
+
+  return (
+    <table className={className} {...props}>
+      {vertical ? (
```
vinodloha: So if table is vertical it doesn't need TBody ?

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,73 @@
...
+        Children.map(children, child => (
+          <tr key={child.props.children}>
+            <th role="columnheader">{child.props.children}</th>
+            {VerticalHeader(child)}
+          </tr>
+        ))
+      ) : (
+        <>
+          <thead>
+            <tr role="row">{Children.map(children, th => Head(th))}</tr>
```
vinodloha: No need for Head here, I think. children could render it like {children}

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,73 @@
...
+          </thead>
+          <tbody>
+            {data.map(item => (
+              <tr key={item.id} role="row">
+                {Children.map(children, td => Body(td, item))}
```
vinodloha: 2 props children, td should be (children, td) => 

Also instead of using Body as function call why don't you apply declarative style?

lib/components/molecules/Table/Table.story.js
```diff
@@ -0,0 +1,53 @@
...
+      <Table.Thead dataMappingKey="name">Dessert (100g serving)</Table.Thead>
+      <Table.Thead dataMappingKey="calories">Calories</Table.Thead>
+      <Table.Thead dataMappingKey="fat">Fat (g</Table.Thead>
+      <Table.Thead dataMappingKey="carbs">Carbs (g)</Table.Thead>
+      <Table.Thead dataMappingKey="protein">Protein (g)</Table.Thead>
+    </Table>
+  ))
+  .add('Table with input box', () => (
+    <Table {...defaultConfig}>
+      <Table.Thead dataMappingKey={() => <InputChoice {...checkboxHeader} />}>
```
vinodloha: Don't like this idea of mixing 2 things in one prop 
> We are using flow so we should not change data type of prop like this (string and function). 

Create a separate prop for () => <InputChoice {...checkboxHeader} /> and also pass the whole row item data in that callback

vinodloha (changes requested):
I think you moved too much away from your earlier state, needs more fixes

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,73 @@
+// @flow
+
+import React, { Children } from 'react';
+import styled from 'styled-components';
+
+import styles from './Table.style';
+import type { Props } from './types';
+
+const Table = ({ children, data, className, vertical, ...props }: Props) => {
```
vinodloha: Table could also have tags like <caption> 

Also look into this for accessibility https://www.w3.org/WAI/tutorials/tables/

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,69 @@
...
+    </td>
+  );
+};
+
+const Table = ({ children, data, className, vertical, ...props }: Props) => {
+  return (
+    <table className={className} {...props}>
+      {vertical ? (
+        <tbody>
+          {Children.map(children, child => (
```
vinodloha: Change Children.map(children, child => to Children.map(item, child =>

And {item} will render the current TH.

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,69 @@
...
+        </tbody>
+      ) : (
+        <>
+          <thead>
+            <tr role="row">{children}</tr>
+          </thead>
+          <tbody>
+            {data.map(item => (
+              <tr key={item.id} role="row">
+                {Children.map(children, child => (
```
vinodloha: Children.map(children, child =>  to 

Also where did you get Children ??? The Capital C in children

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,69 @@
...
+                  <Cell item={item} key={child.id}>
```
vinodloha: I am worried about the handling of key here

vinodloha (changes requested):
Very small things pending into it now

lib/components/molecules/Table/Table.js
ankush29: done

lib/components/molecules/Table/Table.js
ankush29: done.
I imported it from React (React.Children API)

lib/components/molecules/Table/Table.js
ankush29: changed

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,69 @@
...
+const Table = ({ children, data, className, vertical, ...props }: Props) => {
+  return (
+    <table className={className} {...props}>
+      {vertical ? (
+        <tbody>
+          {children.map(items => (
+            <tr key={items.dataMappingKey}>
+              {items}
+              {data.map(item => (
+                <Cell item={item} key={item.id}>
```
vinodloha: key={`${item.id}${items.key}`

ankush29: done

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,69 @@
...
+    </td>
+  );
+};
+
+const Table = ({ children, data, className, vertical, ...props }: Props) => {
+  return (
+    <table className={className} {...props}>
+      {vertical ? (
+        <tbody>
+          {children.map(items => (
```
vinodloha: Do you need to go back to React.Children.map()

lib/components/molecules/Table/types/index.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+import type { Node } from 'react';
+
+export type Props = {
+  className?: string,
+  data: Array<mixed>,
+  children: Node,
+  vertical?: boolean,
+  dataMappingKey: mixed,
```
vinodloha: dataMappingKey should always be a string

lib/components/molecules/Table/types/index.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+import type { Node } from 'react';
+
+export type Props = {
```
vinodloha: Can you rename this to TableProps

lib/components/molecules/Table/Table.js
```diff
@@ -0,0 +1,69 @@
...
+
+import React from 'react';
+import styled from 'styled-components';
+
+import styles from './Table.style';
+import type { Props, HeadProps } from './types';
+
+const Head = ({ children }: HeadProps) => <th>{children}</th>;
+
+const Cell = ({ children, item }: Props) => {
```
vinodloha: This component should need a separate flow definition. How can types be same for 2 components?

ankush29: done

lib/components/molecules/Table/types/index.js
ankush29: done

lib/components/molecules/Table/types/index.js
ankush29: done

lib/components/molecules/Table/Table.js
ankush29: no. converted into flat array using React.Children.toArray()

vinodloha (approved):

PR #175, opened 2019-12-11
WIP | Replace flow with proptype

bhardwajamit0890:


vinodloha:
Before this can we not look to just get rid of flow-typed and just continue using flow? Do a bit of research please

PR #191, opened 2019-12-16
Implemented Icon Component

jainaman23:


lib/components/atoms/Icon/Icon.js
```diff
@@ -0,0 +1,26 @@
...
+import styled from 'styled-components';
+import classnames from 'classnames';
+
+import styles from './Icon.style';
+import type { Props } from './types';
+
+const Icon = ({ className, ...others }: Props): Node => {
+  return (
+    <div role="presentation" className={classnames('icon-wrapper', className)} {...others}>
+      <i className="icon" />
```
vinodloha: We should have a prop that defines Icon name and it can be placed within the body of <i>Close<i> for accessibility

lib/components/atoms/Icon/Icon.style.js
```diff
@@ -0,0 +1,18 @@
+import { css } from 'styled-components';
+
+export default css`
```
vinodloha, line 3: I am not happy with this solution, but for now this will do. 

Let's plan to fix it soon. @jainaman23 assign a ticket to your name

vinodloha (approved):
Work on small fix i have suggested you.

lib/components/atoms/Icon/Icon.js
```diff
@@ -0,0 +1,26 @@
...
+import React from 'react';
+import styled from 'styled-components';
+import classnames from 'classnames';
+
+import styles from './Icon.style';
+import type { Props } from './types';
+
+const Icon = ({ className, title, ...others }: Props): Node => {
+  return (
+    <div role="presentation" className={classnames('icon-wrapper', className)} {...others}>
```
vinodloha: Can you change this icon to span, so that it's inline. Because its a DIV you will not be able to place icon in front of a DIV

vinodloha (approved):

vinodloha (approved):

PR #192, opened 2019-12-17
Implemented Carousel with Declarative Approach

jainaman23:


lib/components/atoms/Button/Button.style.js
```diff
@@ -25,6 +25,21 @@ const ButtonStyles = css`
     outline: none;
   }
 
+  &.circle {
+    border-radius: 50%;
+  }
+
+  &.dots {
```
vinodloha: This should be within the carousel, since this will probably never be needed outside carousel.

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,398 @@
...
+import styled from 'styled-components';
+import classnames from 'classnames';
+
+import stylesCarousel from './Carousel.style';
+import { Props, AnimationKeys } from './types';
+import { ListVanilla } from '../List';
+import Button from '../../atoms/Button';
+import useInterval from './useInterval.hook';
+
+const CarouselContext = React.createContext([]);
```
vinodloha, line 21: Don't use context with-in the molecules.

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,398 @@
...
+ * @param {Array} slides
+ *   all slides
+ *
+ * @return All slides with added slideIndex key as prop.
+ */
+const slideWithIndexKey = slides => {
+  let i = 0;
+  const children = React.Children.map(slides, slide => {
+    i += 1;
+    return React.cloneElement(slide, {
```
vinodloha: This is unnecessary complication. No need to do this if you want to set a incremental id, use a method that returns this value in the default Props.

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,398 @@
...
+  const totalSlidesCount = carouselContent.slides.length;
+  const currentIndex = carouselContent.slideIndex;
+  const setItem = carouselContent.setSlideIndex;
+  const countElements = [];
+
+  for (let id = 1; id <= totalSlidesCount; id += 1) {
+    countElements.push(
+      <li key={id} className={classnames('atomic-item', { active: id === currentIndex })}>
+        <Button onClick={() => setItem(id)} className={classnames('icon', btnType)}>
+          {id}
```
vinodloha: There should be a way to define external component here. If user needs default Dots there is probably no need for an external config but if they want to define the content, provide them with a renderProp.

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,398 @@
...
+ * Provides HTML for next navigation button.
+ *
+ * @param {Object} setIndex
+ *   Change the slide number.
+ * @param {Number} currentIndex
+ *   Current slide number.
+ * @param {Number} slideCount
+ *   Total number of slides.
+ */
+const nextKey = (setIndex, currentIndex, slideCount, children) => {
```
vinodloha: Let's merge this nextKey and prevKey and handle their functional logic through Type. 

use better names get rid of Key its confusing.

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,398 @@
...
+    isAnimationRunning,
+  };
+
+  return (
+    <section className={classnames('carousel-wrapper', className)} aria-label={title}>
+      <CarouselContext.Provider value={value}>
+        {navigation && nextKey(setSlideIndex, slideIndex, slideCount, null)}
+        {children}
+        {navigation && prevKey(setSlideIndex, slideIndex, slideCount, null)}
+        {ariaLiveUpdate(slideIndex, slideCount)}
```
vinodloha: Treat it as component if it returns JSX, Apply the same to rest.

lib/components/molecules/Carousel/Carousel.story.js
```diff
@@ -0,0 +1,57 @@
...
+          <Image src={image2} alt="2.jpg" />
+        </CarouselVanilla.Slide>
+        <CarouselVanilla.Slide>
+          <Image src={image3} alt="3.jpg" />
+        </CarouselVanilla.Slide>
+        <CarouselVanilla.Slide>
+          <Image src={image4} alt="4.jpg" />
+        </CarouselVanilla.Slide>
+      </CarouselVanilla.SlideWrapper>
+      <CarouselVanilla.Dots />
```
vinodloha: Put all the carousel controls in one wrapper. 

<CarouselVanilla.Controls.Dots /><CarouselVanilla.Controls.Next /><CarouselVanilla.Controls.Prev /><CarouselVanilla.Controls.Play />

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,398 @@
...
+    setSlideIndex,
+    setSlidesCount,
+    slideCount,
+    setNavigation,
+    setAnimationRunning,
+    isAnimationRunning,
+  };
+
+  return (
+    <section className={classnames('carousel-wrapper', className)} aria-label={title}>
```
vinodloha: I don't think section is a right tag for carousel. 

--> aria-label={title} title should clarify that its being used as arial label.

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,398 @@
...
+ * @param {Array} children
+ *   Wraps all the slides.
+ *
+ * @return All slides in single wrapper.
+ */
+Carousel.SlideWrapper = ({ children, className }: Props): Node => {
+  const slideContent = useContext(CarouselContext);
+  let currentAnimationStatus = true;
+
+  if (slideContent.slides.length === 0) {
```
vinodloha: This is useless logic please get rid of it.

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,398 @@
...
+    return null;
+  };
+
+  return (
+    <div
+      className={classnames('carousel-section', className)}
+      onMouseEnter={handleMouseEnter}
+      onMouseLeave={handleMouseLeave}
+    >
+      {slideContent.slides}
```
vinodloha: Children should be used here directly.

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,398 @@
...
+ * @return Single slide with wrapper.
+ */
+Carousel.Slide = ({ children, id, className }: Props): Node => {
+  const SlideStatus = useContext(CarouselContext);
+
+  return (
+    <div
+      className={classnames(
+        'slide',
+        { active: id === `slide_${SlideStatus.slideIndex}` },
```
vinodloha: There is no need for such comparison use the ActiveSlide State to make this decision.

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,398 @@
...
+/**
+ * Dots Navigation.
+ *
+ * @param {Element} children
+ *   Will wraps the slide content.
+ *
+ * @return Single slide with wrapper.
+ */
+Carousel.Dots = ({ className, btnType }: Props): Node => {
+  return <div className={classnames('carousel-dots', className)}>{getList(btnType)}</div>;
```
vinodloha: STOP calling functional components like functions use declarative syntax

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,398 @@
...
+};
+
+/**
+ * Next Navigation.
+ */
+Carousel.Next = ({ children }: Nav): Node => {
```
vinodloha: Combine Next and Prev

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+  CarouselProps,
+  AnimationProps,
+  GetListProps,
+  SlideWrapperProps,
+  SlideProps,
+  NavControllerProps,
+} from './types';
+import List from '../List';
+import Button from '../../atoms/Button';
+import useInterval from './useInterval.hook';
```
vinodloha: Can we avoid this naming convention --> ./useInterval.hook

we are placing all hooks at lib\hooks\*

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+ * Increase the slide count.
+ *
+ * @param {Number} type
+ *   Type of navigation control.
+ * @param {Number} setIndex
+ *   Increase the count of slide.
+ * @param {Number} currentIndex
+ *   Show current active slide number.
+ */
+const handleNavController = (type, setIndex, currentIndex) => {
```
vinodloha: For 2 or more parameters use {type, setIndex, currentIndex}

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+ * Increase slide count on key press over next navigation button.
+ *
+ * @param {Object} e
+ *   event parameter
+ * @param {function} setIndex
+ *   increase the slide index.
+ * @param {Number} currentIndex
+ *   Show current active slide number.
+ */
+const handleKeyPress = (e, setIndex, currentIndex) => {
```
vinodloha: You don't have to do this the prop type on Nav control could tell the handleNavController if its next or prev. Checking this on Event is slow and jQuery way of doing it. 

'next', and 'prev', could be the value of it.

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+  return null;
+};
+
+/**
+ * Provides HTML for navigation button.
+ *
+ * @param {Object} type
+ *   type of navigation controller.
+ */
+const NavController = ({ type, children, ...others }: NavControllerProps): Node => {
```
vinodloha: This type would be set by the developer and you can define a local click handler that reads type and calls parent onclick handler from it

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+  const { setSlideIndex, slideIndex, slides } = useContext(CarouselContext);
+
+  return (
+    <div className={classnames('nav-action', type)} {...others}>
```
vinodloha: Whats the point of putting {...others} on DIV what kind of values you plan to set on Div

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+      <Button
+        onClick={() => handleNavController(type, setSlideIndex, slideIndex)}
+        onKeyUp={e => handleKeyPress(e, setSlideIndex, slideIndex)}
+        disabled={type === 'next' ? slideIndex === slides.length - 1 : slideIndex === 0}
+        aria-disabled={type === 'next' ? slideIndex === slides.length - 1 : slideIndex === 0}
```
vinodloha: Don't think there is a need for aria-disabled. 

https://stackoverflow.com/questions/38059140/when-to-use-the-disabled-attribute-vs-the-aria-disabled-attribute-for-html-eleme

Also don't duplicate expressions like you did in terms of finding a value, no harm in declaring a variable and using it

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+      >
+        {children || type}
```
vinodloha: We can't use type as content and children should be fallback in such cases. Preference should be given to props user is setting. Use an explicit label.

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+ *   Active slide number.
+ * @param {Number} slides
+ *   All slides.
+ */
+const AriaLiveUpdate = () => {
+  const { slideIndex, slides } = useContext(CarouselContext);
+
+  return (
+    <div className="hidden" aria-live="polite" aria-atomic="true" tabIndex="-1">
+      {`Item ${slideIndex + 1} of ${slides.length}`}
```
vinodloha: Again no hard coding of text, our components should always be localisation friendly. 

This accessibility should be backed on Carousel Indicators because they are the ones  that are showing this info on UI.

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+      <AriaLiveUpdate />
+      <Animation {...props} />
+    </div>
+  );
+};
+
+/**
+ * Play and Pause icon to handle slide animation.
+ */
+const Animation = ({ className, playLabel, pauseLabel }: AnimationProps): Node => {
```
vinodloha: Animation --> AnimationControl

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+  const { setAnimationRunning, isAnimationRunning } = useContext(CarouselContext);
+
+  const handleAnimation = () => {
+    setAnimationRunning(!isAnimationRunning);
+  };
+
+  return (
+    <div className={classnames('animation-panel', className)} aria-label="animation-controller">
```
vinodloha: aria-label="animation-controller" is machine reachable text can't be hard coded and can't be kabab-case

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+      <Button onClick={handleAnimation}>{isAnimationRunning ? pauseLabel : playLabel}</Button>
```
vinodloha: Do we need text or Icons or both for these controls?

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+      setAnimationRunning(!currentAnimationStatus);
+    }
+
+    return null;
+  };
+
+  return (
+    <div
+      className={classnames('carousel-section', className)}
+      onMouseEnter={handleMouseEnter}
```
vinodloha, line 301: Shouldn't you attach this events only when carousel is AutoPlay in first place? Else null

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+      onMouseLeave={handleMouseLeave}
+      {...others}
+    >
+      {React.Children.map(children, (child, index) => {
+        return React.cloneElement(child, { id: index });
```
vinodloha: Is this ID only needed for accessibility?

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+ * Dots Navigation.
+ *
+ * @param {Element} children
+ *   Will wraps the slide content.
+ *
+ * @return Single slide with wrapper.
+ */
+Carousel.Dots = ({ className, btnType, thumbnails }: GetListProps): Node => {
+  return (
+    <div className={classnames('carousel-dots', className)}>
```
vinodloha: Why everything needs a div Wrapper. Didn't we discussed that we can place all controls on one container and then we can position it with CSS?

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+
+/**
+ * Dots Navigation.
+ *
+ * @param {Element} children
+ *   Will wraps the slide content.
+ *
+ * @return Single slide with wrapper.
+ */
+Carousel.Dots = ({ className, btnType, thumbnails }: GetListProps): Node => {
```
vinodloha: What the role of btnType and thumbnails here?

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+
+/**
+ * Next Navigation.
+ */
+Carousel.Next = ({ ...props }) => <NavController type="next" {...props} />;
+
+/**
+ * Previous Navigation.
+ */
+Carousel.Prev = ({ ...props }) => <NavController type="prev" {...props} />;
```
vinodloha: Nice, you already have type and clickable controls here so not sure what all that logic of determining button type was about.

lib/components/molecules/Carousel/Carousel.js
vinodloha: Should be tested independently and exported independently as well.

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+import useInterval from './useInterval.hook';
+import CarouselContext from './CarouselContext';
+
+/**
+ * Rendering carousel dots
+ *
+ * @return HTML
+ *   Wrapper with UL | OL
+ */
+const GetList = ({ btnType, thumbnails }: GetListProps): Node => {
```
vinodloha: I don't think we need to maintain variations from compound components. Carousel will have variation not each compound component. 

If you need control on lets say Button it should make Use of Button Components's API

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+  const { slideIndex, setSlideIndex, slides } = useContext(CarouselContext);
+  const allSlides = [];
+
+  for (let id = 0; id < slides.length; id += 1) {
+    allSlides.push(
+      <li key={id} className={classnames('item', { active: id === slideIndex })}>
+        <Button onClick={() => setSlideIndex(id)} className={classnames('icon', btnType)}>
+          {thumbnails[id] || id + 1}
```
vinodloha: Instead of expecting user to pass an array lets give then RenderProp like renderSlideThumb on Carousel, so that they can set the content of the Thumb as per the slide indexes. 

This way you don't have to expose ListWrapper and you can place it yourself. Also we can get rid of extra props. 

If  renderSlideThumb for Thumb is defined then:
1. Please the list below the carousel. 
2. Show indicators as Square box rather than Circles. 
3. Button will be Transparent.

Else 
1. Place Dots on slides, 
2. Keep then circular.

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+import CarouselContext from './CarouselContext';
+
+/**
+ * Rendering carousel dots
+ *
+ * @return HTML
+ *   Wrapper with UL | OL
+ */
+const GetList = ({ btnType, thumbnails }: GetListProps): Node => {
+  const { slideIndex, setSlideIndex, slides } = useContext(CarouselContext);
```
vinodloha: slideIndex --> activeSlideIndex

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+ *
+ * @param {Number} type
+ *   Type of navigation control.
+ * @param {Number} setIndex
+ *   Increase the count of slide.
+ * @param {Number} currentIndex
+ *   Show current active slide number.
+ */
+const handleNavController = (type, setIndex, currentIndex) => {
+  setIndex(type === 'next' ? currentIndex + 1 : currentIndex - 1);
```
vinodloha: setIndex --> setSlideIndex
Why we can't use setSlideIndex directly on the NavControl?

lib/components/molecules/Carousel/Carousel.js
vinodloha: type --> navType

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+};
+
+/**
+ * Carousel Controls.
+ */
+const CarouselControls = ({ ...props }) => {
+  return (
+    <div className="carousel-controls">
+      <NavController type="next" />
+      <NavController type="prev" />
```
vinodloha: Take Prev at the Last.

lib/components/molecules/Carousel/Carousel.js
vinodloha: pauseLabel : playLabel should be named better so we know they are render props.

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+
+  useInterval(
+    () => {
+      const currentIndex = slideIndex !== slides.length - 1 ? slideIndex : -1;
+      setSlideIndex(currentIndex + 1);
+    },
+    isAnimationRunning ? slideTransitionTime : null
+  );
+
+  const value = {
```
vinodloha: value --> carouselState

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+    setSlidesContent,
+    isAnimationRunning,
+    setAnimationRunning,
+  };
+
+  return (
+    <section className={classnames('carousel-wrapper', className)} {...others}>
+      <CarouselContext.Provider value={value}>
+        {children}
+        {controls && <CarouselControls />}
```
vinodloha: Nice

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+
+/**
+ * Slide Wrapper.
+ *
+ * @param {Array} children
+ *   Wraps all the slides.
+ *
+ * @return All slides in single wrapper.
+ */
+Carousel.SlideWrapper = ({ children, className, ...others }: SlideWrapperProps): Node => {
```
vinodloha, line 260: We need responsive support: With current index we ca't achieve responsive carousel 

Hint --> Flim Strip

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,330 @@
...
+ */
+Carousel.AnimationController = ({ ...props }) => <Animation {...props} />;
+
+/**
+ * Carousel Controls
+ */
+Carousel.Controls = ({ ...props }) => <CarouselControls {...props} />;
+
+Animation.defaultProps = {
+  playLabel: 'play',
```
vinodloha: Display test should never be part of default Props.

lib/components/molecules/Carousel/Carousel.story.js
```diff
@@ -0,0 +1,61 @@
...
+        </CarouselVanilla.Slide>
+        <CarouselVanilla.Slide>
+          <Image src={image4} alt="4.jpg" />
+        </CarouselVanilla.Slide>
+      </CarouselVanilla.SlideWrapper>
+      <CarouselVanilla.Controls />
+    </CarouselVanilla>
+  ))
+  .add('Carousel', () => (
+    <Carousel {...defaultConfig}>
```
vinodloha, line 37: We must document both the variations with Controls and without so we know how Carousel can be used. Also the responsive uses too.

lib/components/molecules/Carousel/CarouselContext.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+const CarouselContext = React.createContext([]);
```
vinodloha: Lets keep this with-in the component only.

vinodloha (changes requested):
Need further improvements

lib/components/molecules/Carousel/Carousel.js
jainaman23: We agreed upon this.

lib/components/molecules/Carousel/Carousel.js
jainaman23: It was not working like this

lib/components/molecules/Carousel/Carousel.js
jainaman23: Responsive Implementation has been added.

lib/components/molecules/Carousel/Carousel.story.js
jainaman23: Done

lib/components/atoms/Button/Button.js
```diff
@@ -14,6 +14,7 @@ type Props = {
   primary?: boolean,
   secondary?: boolean,
   tertiary?: boolean,
+  bullets?: boolean,
   disabled?: boolean,
   variation?: String,
```
vinodloha: Why you can't handle it via variation property? Why increase the API

lib/components/atoms/Button/Button.style.js
```diff
@@ -38,6 +46,10 @@ const ButtonStyles = css`
     ${props.theme.colors.CTA_TRANSPARENT}
   `};
 
+  &.circle {
```
vinodloha: How is this variation diff than bullets and why its needed?

lib/components/molecules/Carousel/Carousel.story.js
```diff
@@ -0,0 +1,107 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import defaultConfig, { responsiveConfig, vanillaConfig, centerModeConfig } from './Carousel.mock';
+import Image from '../../atoms/Image';
+import image1 from '../../../static/images/dummies/1280.png';
+import image2 from '../../../static/images/dummies/960.png';
```
vinodloha: No need to import, just put the path in src

lib/components/molecules/Carousel/Carousel.js
```diff
@@ -0,0 +1,398 @@
...
+  prevLabel: 'prev',
+};
+
+AnimationControl.defaultProps = {
+  renderPlay: () => 'play',
+  renderPause: () => 'pause',
+};
+
+Carousel.Dots.defaultProps = {
+  thumbnails: [],
```
vinodloha: This Prop is not used any where? Why does this like exists then?

vinodloha:
Screenshot 2020-01-06 at 3 09 04 PM

Getting this error when checking responsive carousel in mobile viewport 

Also, you must export the Carousel and CarouselVanilla out of Component library. (lib/index.js)

vinodloha (changes requested):
Needs few fixes

vinodloha (approved):
Great quality code @jainaman23 Finally :)

Issue #85, opened 2019-10-17
Add Tag/Pill Component


manishekhawat:
**Issue Statement**

Create a generic Tag component

**Summary**

Create an independent react component with following requirement considerations:

1. A simple tag component
2. It will hold its independent state as well as can expose an API to control the dismiss or remove functionality

**Screenshots**

<img width="942" alt="Screenshot 2019-10-17 at 10 17 00 PM" src="https://user-images.githubusercontent.com/8122967/67029916-e470db80-f12b-11e9-836a-d3e1b8ae6e95.png">

<img width="304" alt="Screenshot 2019-10-17 at 10 43 41 PM" src="https://user-images.githubusercontent.com/8122967/67031663-a07fd580-f12f-11e9-8f9e-1febb4fc4ea2.png">

PR #193, opened 2019-12-19
feat(#85): tag component

achaljain:
- new atom component - tag
- variations: default, removable
- renders text and remove button
- provides api for parent on remove


lib/components/atoms/Button/Button.js
```diff
@@ -15,6 +15,7 @@ type Props = {
   secondary?: boolean,
   tertiary?: boolean,
   disabled?: boolean,
+  icon?: boolean,
```
vinodloha: I don't think it's uses clear from the prop name. 

We should define a prop --> variation and its value could be 'icon-only'

Later we can plan to move other variations like Primary/Secondly under this.

lib/components/atoms/Button/Button.style.js
```diff
@@ -19,6 +19,14 @@ const ButtonStyles = css`
     ${props.theme.border.CTA_TERTIARY};
   `
       : ``};
+  ${props =>
+    props.icon &&
+    `
+    min-height: initial;
```
vinodloha: No need for **min-height: initial;** Let all button be of same height. You can overwrite this in your component as its need of your component to not have default height. Having CTA size 40x40 is a good accessibility practice. 

Secondly, If you keep top and bottom padding 0 how will icon get aligned in middle?

lib/styles/themes/base/def/colors.js
```diff
@@ -22,6 +22,14 @@ export default {
       color: var(--text-inverse);
     }
   `,
+  CTA_ICON: `
+    background-color: var(--white);
+    color: var(--black);
+    &:hover, &:active{
+      background-color: var(--white);
+      color: var(--black);
```
vinodloha: Since we are using SVG icons to change the color we use Fill property not color

lib/styles/themes/base/def/colors.js
```diff
@@ -67,4 +75,7 @@ export default {
   TEXT_LIGHT: `
     color: var(--text-light);
   `,
+  BG_DEFAULT: `
```
vinodloha: Wrong variable it should be more specific to what exactly does this affect in UI.

lib/components/atoms/Tag/types/index.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+import type { Node } from 'react';
+
+export type Props = {
```
vinodloha: Props --> TagProps

lib/components/atoms/Tag/types/index.js
```diff
@@ -0,0 +1,9 @@
...
+  children: Node,
+  className?: String,
+  variation?: String,
+  onRemove?: Function,
```
vinodloha, line 9: I think you will need a name property too that identifies which tag was removed.May be keep a hidden input inside. Thoughts?

lib/components/atoms/Tag/Tag.style.js
```diff
@@ -0,0 +1,18 @@
...
+
+  display: inline-flex;
+  align-items: center;
+
+  ${props => `
+    ${props.theme.border.TAG};
+    ${props.theme.colors.TEXT_DEFAULT};
+    ${props.theme.colors.BG_DEFAULT};
+  `}
+
```
vinodloha, line 13: You style overwrites on standard button should come here.

lib/components/atoms/Tag/Tag.style.js
```diff
@@ -0,0 +1,18 @@
...
+  .text-content {
+    padding: ${props => props.theme.spacing.UNIT_1};
+  }
```
vinodloha, line 16: Make use of inheritedStyles prop that rest of the components use to allow CSS overwrites

lib/components/atoms/Tag/Tag.js
```diff
@@ -0,0 +1,50 @@
...
+import styles from './Tag.style';
+import type { Props } from './types';
+
+const Tag = ({ className, children, variation, onRemove }: Props): Node => {
+  const [showTag, setShowTag] = React.useState(true);
+
+  const removeTag = e => {
+    setShowTag(false);
+    if (onRemove) {
+      onRemove(e);
```
vinodloha: Call back could also send which tag was removed. Else the remove won't mean much

lib/components/atoms/Tag/Tag.js
```diff
@@ -0,0 +1,50 @@
...
+import React from 'react';
+import styled from 'styled-components';
+import classnames from 'classnames';
+
+import Button from '../Button';
+import styles from './Tag.style';
+import type { Props } from './types';
+
+const Tag = ({ className, children, variation, onRemove }: Props): Node => {
+  const [showTag, setShowTag] = React.useState(true);
```
vinodloha: showTag --> isVisible

lib/components/atoms/Tag/Tag.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+  };
+
+  return (
+    <>
+      {showTag && (
+        <span className={classnames(className)}>
+          <span className="text-content">{children}</span>
+          {variation === 'removable' && (
+            <Button icon className="remove-tag" onClick={removeTag} ariaLabel="Remove Tag">
```
vinodloha: This className="remove-tag" doesn't mean much in react. 

No hard coding of Test please even if its for Aria labels

lib/components/atoms/Tag/Tag.js
```diff
@@ -0,0 +1,50 @@
...
+    setShowTag(false);
+    if (onRemove) {
+      onRemove(e);
+    }
+  };
+
+  return (
+    <>
+      {showTag && (
+        <span className={classnames(className)}>
```
vinodloha: Not very happy with the choice of tags, please explore and research also accessibility handling needs to be better. For example, how will a disabled user will know if a tag is selected or getting removed?

vinodloha (changes requested):
Needs given improvements.

lib/components/atoms/Button/tests/__snapshots__/Button.style.test.js.snap
```diff
@@ -104,6 +110,13 @@ exports[`Button should render correctly 1`] = `
         "CTA_ERROR": "
     background-color:var(--bg-cta-error);
     color:var(--text-cta-error);",
+        "CTA_ICON": "
+    background-color: var(--white);
```
vinodloha: I think the background should be transparent

lib/components/atoms/Tag/Tag.js
```diff
@@ -0,0 +1,64 @@
...
+    setIsVisible(false);
+    if (onRemove) {
+      onRemove(e, tagName);
+    }
+  };
+
+  return (
+    <>
+      {isVisible && (
+        <span id={tagName} className={classnames(className)}>
```
vinodloha: Is there a role that can be used here to tell what kind of element it is?

lib/components/atoms/Tag/tests/Tag.test.js
```diff
@@ -0,0 +1,40 @@
...
+  });
+
+  test('should render provided text', () => {
+    expect(TagComponent.text()).toEqual('Test');
+  });
+
+  test('should control remove button availability through prop', () => {
+    expect(TagComponent.exists('Button[variation="icon-only"]')).toEqual(false);
+    TagComponent.setProps({ variation: 'removable' });
+    console.log('Tagcomponent === ', TagComponent.debug());
```
vinodloha: Remove logs

vinodloha (approved):

PR #197, opened 2019-12-23
StickyHeader | Review comments incorporated

Vipul6:


utils/util.js
```diff
@@ -0,0 +1,7 @@
+const useHandleScroll = (elemRef, setterFunc) => {
```
vinodloha: The name of the hook is ambiguous and also not placed correctly. We are keeping hooks at lib/hooks/*

lib/index.js
```diff
@@ -79,4 +80,6 @@ export {
   TypeaheadVanilla,
   Video,
   VideoVanilla,
+  StickyHeader,
```
vinodloha: Follow the alphabetical order

lib/components/molecules/StickyHeader/StickyHeader.js
```diff
@@ -0,0 +1,31 @@
...
+  useEffect(() => {
+    window.addEventListener('scroll', () => useHandleScroll(wrapperRef, setSticky));
+    return () => {
+      window.removeEventListener('scroll', useHandleScroll);
+    };
+  });
+
+  return (
+    <div aria-label="Sticky Header" className={`${isSticky ? className : ''}`} ref={wrapperRef}>
+      <header>{children}</header>
```
vinodloha: @Vipul6 I don't think you have fixed the original problem i had highlighted. This component is causing a flicker when it gets sticky and also hiding the page content below itself.

vinodloha (changes requested):
This still needs few fixes

utils/util.js
Vipul6: It has been taken care.

lib/index.js
Vipul6: It has been taken care.

lib/components/molecules/StickyHeader/StickyHeader.js
Vipul6: Fixed.

lib/components/molecules/StickyHeader/StickyHeader.js
```diff
@@ -0,0 +1,36 @@
...
+  useEffect(() => {
+    window.addEventListener('scroll', () => useScroll(wrapperRef, setSticky));
+    return () => {
+      window.removeEventListener('scroll', useScroll);
+    };
+  });
+
+  return (
+    <Wrapper
+      aria-label="Sticky Header"
```
vinodloha: Don't hard code any string in the component, take them through props please.

lib/components/molecules/StickyHeader/StickyHeader.story.js
```diff
@@ -0,0 +1,40 @@
...
+        </StickyHeader>
+        <p>
+          Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has
+          been the industrys
+        </p>
+        <p>
+          Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has
+          been the industrys
+        </p>
+        <StickyHeader>
```
vinodloha: Tested this, still not working as expected. When this instance gets sticky it should not overlap the above header unless explicitly set to do so. You had to create a prop that takes sticky position value. 

We had discussed this in quite a detail.

lib/components/molecules/StickyHeader/StickyHeader.js
```diff
@@ -0,0 +1,36 @@
...
+      className={`${isSticky ? className : 'no-sticky'}`}
+      ref={wrapperRef}
+      elmRef={wrapperRef}
+    >
+      <header>{children}</header>
```
vinodloha: If there could be multiple Header tag on page, it should be wrapped in Section

lib/components/molecules/StickyHeader/StickyHeader.style.js
```diff
@@ -0,0 +1,16 @@
...
+  header {
+    position: fixed;
+    top: 0;
+    width: 100vw;
+  }
+`;
+
+export const Wrapper = styled.div`
+  position: relative;
+  height: ${props => props.elmRef.current != null && `${props.elmRef.current.offsetHeight}px`};
```
vinodloha: This should be set inline on the element.

lib/hooks/useScroll.js
```diff
@@ -0,0 +1,7 @@
+const useScroll = (elemRef, setterFunc) => {
```
vinodloha: It should be throttled/debounced, else it will execute on each pixel

vinodloha (changes requested):
Needs more improvement.

lib/components/molecules/StickyHeader/StickyHeader.js
aashutosh-prakash: aria-label wont work on non-interactive elements like Wrapper (originally a div tag).
Removed it from code.

aashutosh-prakash:
Raised new PR with commit lint fixes.

PR #198, opened 2019-12-23
Feature: File upload Functionality changes pushed

salilkapoor:
Changes related to file upload functionality.


lib/components/molecules/Form/Form.story.js
```diff
@@ -24,6 +26,9 @@ storiesOf('Molecules/Form', module)
         <Form.Select {...userFormSelectProps} />
         <Form.Input {...userFormNameInputProps} />
         <Form.Input {...userFormEmailProps} />
+        <Form.FieldFile {...fileMultipleInputProps} />
```
vinodloha, line 29: @salilkapoor you should also document these independently, just like FieldTextArea 

also please put few headings or paragraphs between content of this form. When i ran storybook it looks so chaotic

lib/hooks/useDragNDrop.js
```diff
@@ -0,0 +1,58 @@
...
+  };
+
+  const getHighlightClass = className => {
+    const styleCls = classNames(className, {
+      highlight,
+    });
+    return styleCls;
+  };
+
+  return {
```
vinodloha, line 46: @salilkapoor is there a way someone could filter out files being uploaded through drag and drop and selection? I think it input[type="file"] its easy to set file types but drag and drop would need some mechanics for it. 

Like if the drop only supports .png then other files should be excluded with a message to user. 

Also there should be a way to set upload file-size limit, single or collective.

Sorry for not defining it as par of  requirement.

lib/hooks/useDragNDrop.js
```diff
@@ -0,0 +1,58 @@
...
+  const onDrop = event => {
+    event.preventDefault();
+    const { files } = event.dataTransfer;
+    const array = fileListToArray(files);
+    return array;
+  };
+
+  const onFilesAdded = evt => {
+    const { files } = evt.target;
+    const array = fileListToArray(files);
```
vinodloha, line 35: You could return from this line itself 

```javascript
return fileListToArray(files);
```

lib/hooks/useDragNDrop.js
```diff
@@ -0,0 +1,58 @@
...
+    setHighlight(true);
+  };
+
+  const onDragLeave = e => {
+    e.preventDefault();
+    setHighlight(false);
+  };
+
+  const onDrop = event => {
+    event.preventDefault();
```
vinodloha, line 27: Are you sure you need to do event.preventDefault(); ?? I am unable to drop any file on my machine

vinodloha (approved):
Looks very good expect for small issues and few enhancements

PR #202, opened 2019-12-27
Toast-Icon-Component-Dependency-Resolved

SakshiGrg:
Toast-Icon-Component-Dependency-Resolved Review Required


lib/styles/themes/base/def/icons.js
```diff
@@ -1,3 +1,15 @@
 import close from '../icons/close-black.svg';
+import closewhite from '../icons/close.svg';
```
vinodloha, line 2: can we not use the black icon only? We should be able to change SVG color with fill property

SakshiGrg: Right now we can not use SVG color with fill property because the Icon Component code is not SVG Dependent. 
Icon is coming as per Background-image property as per icon(id) passed into it as a prop.
I have some concerns regarding the same. Me and Saurav did some R&D on that. 
As per current code, we can use filter css property for color change.

lib/components/molecules/Toast/Toast.js
```diff
@@ -42,7 +43,7 @@ const Toast = ({ children, autoDismissTimeout, className, isVisible }: Props): N
       {showToast && (
         <div className={classnames('toast', className)}>
           <Button className="toast-close" onClick={close}>
-            <img src="/images/close.svg" alt="close button" />
+            <Icon id="closewhite" width="12px" height="12px" title="close icon" />
```
vinodloha, line 46: Don't Hard code the text. title="close icon"

vinodloha (approved):

PR #205, opened 2019-12-27
Feature kd pagination

desai-koustubh:
Hi Vinod,
Pagination component is right now satisfying two counts:
1. I can use it with a table & list and any other markup
2. Managed area with render props:

Problem: 
1. Not appropriate for table markup since buttons are also shown up as a part of table.

Q. Do I need to make Pagination a HOC?
Q. I am doing something fundamentally wrong. Please give me hints or keywords to refactor (Not the logic. I will figure out the logic). Is that fine?


lib/components/molecules/Pagination/Pagination.js
```diff
@@ -0,0 +1,52 @@
...
+import type { Node } from './types';
+import { noop } from '@babel/types';
+
+type Props = {
+  size: number,
+  render: Function,
+  items: Array,
+  className: Array | string,
+};
+const buttons = (n, component) => Array.from('x'.repeat(n), (_, s) => component(s + 1));
```
vinodloha, line 20: Array.from is not yet supported by IE 11, whats the fallback for it?

lib/components/molecules/Pagination/Pagination.js
```diff
@@ -0,0 +1,52 @@
...
+const Pagination = ({ size, render, items }: Props): Node => {
+  const [current, setCurrentPage] = useState(1);
+  const pages = Math.ceil(items.length / size);
+  const button = n => (
+    <button key={n} onClick={() => setCurrentPage(n)}>
+      Page {n}
+    </button>
+  );
+  const pageButtons = pages > 1 ? buttons(pages, button) : '';
```
vinodloha, line 29: This is too much of imperative programming, React recommendation is for Declarative programming over imperative

lib/components/molecules/Pagination/Pagination.js
```diff
@@ -0,0 +1,52 @@
...
+  const start = (current - 1) * size;
+  const end = start + size;
+  items = items.slice(start, end);
```
vinodloha, line 32: You are not treating list in immutable way. Where you will get the next set of data if you overwrite the original source in the first pagination action itself.

lib/components/molecules/Pagination/Pagination.js
```diff
@@ -0,0 +1,52 @@
...
+  render: Function,
+  items: Array,
+  className: Array | string,
+};
+const buttons = (n, component) => Array.from('x'.repeat(n), (_, s) => component(s + 1));
+const Pagination = ({ size, render, items }: Props): Node => {
+  const [current, setCurrentPage] = useState(1);
+  const pages = Math.ceil(items.length / size);
+  const button = n => (
+    <button key={n} onClick={() => setCurrentPage(n)}>
```
vinodloha, line 25: This onClick={() => setCurrentPage(n)} forms a closure with-in the component and quite bad for React from performance stand point

lib/components/molecules/Pagination/Pagination.js
```diff
@@ -0,0 +1,52 @@
...
+      Page {n}
+    </button>
+  );
+  const pageButtons = pages > 1 ? buttons(pages, button) : '';
+  const start = (current - 1) * size;
+  const end = start + size;
+  items = items.slice(start, end);
+  return (
+    <div>
+      <>{render(items)}</>
```
vinodloha, line 35: The semantic markup is quite Poor

lib/components/molecules/Pagination/tests/Pagination.test.js
```diff
@@ -0,0 +1,15 @@
...
+
+import Pagination from '../index';
+
+describe('<Pagination />', () => {
+  let PaginationComponent = '';
+  beforeEach(() => {
+    PaginationComponent = shallow(<Pagination>Test</Pagination>);
+  });
+
+  test('should render correctly', () => {
```
vinodloha, line 12: We plan to do TDD so we should attempt to write code in that format

lib/components/molecules/Pagination/Pagination.js
```diff
@@ -0,0 +1,52 @@
...
+  size: number,
+  render: Function,
+  items: Array,
+  className: Array | string,
+};
+const buttons = (n, component) => Array.from('x'.repeat(n), (_, s) => component(s + 1));
+const Pagination = ({ size, render, items }: Props): Node => {
+  const [current, setCurrentPage] = useState(1);
+  const pages = Math.ceil(items.length / size);
+  const button = n => (
```
vinodloha, line 24: All components should be named in Title case

lib/components/molecules/Pagination/Pagination.js
desai-koustubh: Can use a core.js polyfill

PR #207, opened 2020-01-06
Feature/global search box

imsauravsingh:
-> functional implementation is done
-> Found few issues on typeahead where I am working
-> Test cases are still pending (Pending...)
-> Icon component needs to implement. (it wasn't working as per expectation)


stylelint.config.js
```diff
@@ -1,6 +1,6 @@
 const strictRules = {
   'selector-max-specificity': [
-    '0,3,0',
+    '0,4,0',
```
vinodloha: This too much deep selector, we should avoid such scenarios.

lib/components/molecules/Search/SearchList/SearchList.js
```diff
@@ -0,0 +1,33 @@
+// @flow
+import React, { Node } from 'react';
+import styled from 'styled-components';
+
+import type { Props } from './types';
+import styles from './SearchList.style';
+import List from '../../List';
+
+const SearchList = ({ dataLists, renderList }: Props): Node => {
```
vinodloha: I think this is already available in typeahead

vinodloha:
Screenshot 2020-01-07 at 9 01 31 AM

UI is broken for search results.

vinodloha (changes requested):
Reported few issues.

lib/components/molecules/Search/Search.js
```diff
@@ -0,0 +1,110 @@
...
+  const onSpeechClick = () => onSpeechListenHandler();
+
+  return (
+    <div className={classnames(styles.default, className)}>
+      <div className="search-box">
+        <Typeahead
+          className={classnames('search-txt', inputProps.className)}
+          id="typeahead"
+          name="wordsearch"
+          placeholder="Type any word"
```
vinodloha, line 63: @imsauravsingh please don't hard code any text, it should be prop driven

lib/components/molecules/Search/Search.js
```diff
@@ -0,0 +1,110 @@
...
+  });
+
+  const onSpeechClick = () => onSpeechListenHandler();
+
+  return (
+    <div className={classnames(styles.default, className)}>
+      <div className="search-box">
+        <Typeahead
+          className={classnames('search-txt', inputProps.className)}
+          id="typeahead"
```
vinodloha, line 61: There should be a way to set the ID and Name from outside. This can be a fallback default value

lib/components/molecules/Search/Search.js
```diff
@@ -0,0 +1,110 @@
...
+    handleSpeechEnd,
+  });
+
+  const onSpeechClick = () => onSpeechListenHandler();
+
+  return (
+    <div className={classnames(styles.default, className)}>
+      <div className="search-box">
+        <Typeahead
+          className={classnames('search-txt', inputProps.className)}
```
vinodloha, line 60: search-txt --> search-text

lib/components/molecules/Search/Search.js
```diff
@@ -0,0 +1,110 @@
...
+          minItemsInList={3}
+        />
+        <div className="search-button">
+          <Button>
+            {browserSupported &&
+              (isListening ? (
+                <i className="search-icon search-loader" />
+              ) : (
+                // eslint-disable-next-line jsx-a11y/control-has-associated-label
+                <i
```
vinodloha, line 90: Why we can't use the button component here?

lib/components/molecules/Search/Search.js
```diff
@@ -0,0 +1,110 @@
...
+          getResults={() => {
+            customFilter(searchString);
+          }}
+          getSelectedResultValue={value => {
+            setSearchString(value);
+            getSelectedResultValue(value);
+          }}
+          renderCustomResultsView={item => <p>{item}</p>}
+          isFieldLabel="true"
+          label="search word"
```
vinodloha, line 79: Same here hard coded labels

vinodloha (approved):

PR #208, opened 2020-01-08
Feature/progress indicator 179

SakshiGrg:
Review Required


lib/components/molecules/ProgressIndicator/ProgressIndicator.js
```diff
@@ -0,0 +1,98 @@
...
+    });
+  };
+
+  /*
+   * calling renderChildren function outside the return statement in order to prevent it from calling everytime during runtime in React Component
+   */
+  const renderChild = renderChildren();
+  return (
+    <>
+      <div className={classnames('progress-indicator-container', className)}>
```
vinodloha, line 67: We need to use HTML5 progress

PR #213, opened 2020-03-24
added react/svg module  for creading svg icon in icon component

anilvermaspeaks:
package added for svg is - react-svg

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,6 +137,7 @@
     "react-app-polyfill": "^1.0.4",
     "react-dom": "^16.10.0",
     "react-modal": "^3.6.1",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -137,6 +137,7 @@
     "react-app-polyfill": "^1.0.4",
     "react-dom": "^16.10.0",
     "react-modal": "^3.6.1",
+    "react-svg": "^11.0.15",
```
vinodloha, line 140: @anilcs240 This package is ok 250KB size and it gets used at run time. The solution we needed was something that works on build time and converts SVG to JS files without adding anything extra to bundle size.

PR #214, opened 2020-03-24
WIP: fix(formik-to-react-hook-form): formik to react hook form

ankush29:
update input type only


lib/components/molecules/Form/FieldInput/FieldInput.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 import React from 'react';
 import type { Node } from 'react';
-import { Field, ErrorMessage } from 'formik';
+// import { Field, ErrorMessage } from 'formik';
```
vinodloha: Don't keep it commented

package.json
```diff
@@ -128,15 +128,17 @@
...
     "flexboxgrid": "^6.3.1",
     "formik": "^2.0.6",
     "lodash": "^4.17.15",
     "normalize.css": "^8.0.0",
     "react": "^16.10.0",
     "react-app-polyfill": "^1.0.4",
     "react-dom": "^16.10.0",
+    "react-hook-form": "^5.1.3",
     "react-modal": "^3.6.1",
+    "regenerator-runtime": "^0.13.5",
```
vinodloha: Why do we need this package?

lib/components/molecules/Form/Form.js
```diff
@@ -15,95 +16,67 @@ import FieldFileInput from './FieldFileInput';
...
 
-  static Checkbox = FieldCheckbox;
-
-  static Field = Field;
-
-  static Input = FieldInput;
+Form.Checkbox = FieldCheckbox;
 
-  static RadioGroup = FieldRadioGroup;
+Form.Input = FieldInput;
```
vinodloha, line 58: These static exports should be moved out of component body.

lib/components/molecules/Form/Form.js
```diff
@@ -15,95 +16,67 @@ import FieldFileInput from './FieldFileInput';
...
+  validationSchema,
+  yupValidationSchema,
+}: Props) => {
+  const getSchema = () => {
     if (yupValidationSchema) return yupValidationSchema;
     return validationSchema ? buildSchema(validationSchema) : null;
   };
+  const { handleSubmit, register, errors } = useForm({
+    mode: validate,
+    validationSchema: getSchema(yupValidationSchema, validationSchema),
```
vinodloha, line 41: Do we still want to continue using Yup for validation? This was a dependency in formik may be we can avoid it. Thoughts?

ankush29: we can also do the validation using yup in react-hook-form

PR #224, opened 2020-03-31
React hook form new

ankush29:


lib/components/atoms/Input/Input.js
```diff
@@ -20,20 +20,21 @@ const Input = React.forwardRef<Props, HTMLInputElement>(
...
+    const propArray = {
+      id,
+      ref,
+      'aria-label': ariaLabel,
+      className,
+      name,
+      type,
+      placeholder,
+      'data-locator': elementLocator || `input-${type}-${name}`,
+      ...others,
```
vinodloha, line 33: @ankush29 Please look at Achal how he has fixed flow errors. I think one of the main issue here i see it about uses of ...others prop

I think as per flow this should be first prop and specific ones should be set later. @achaljain can help please connect

